PR #30, opened 2018-09-22
Featrue/optimize discovery

xiejiashuai:
优化了一些代码结构，望采纳。


fangjian0423:
nacos discovery has been refactored.

PR #57, opened 2018-10-24
fix sentinel bug and add a new module spring-cloud-alibaba-sentinel-datasource in branch 1.x

fangjian0423:
fix sentinel bug and add a new module spring-cloud-alibaba-sentinel-datasource separate from spring-cloud-alibaba-sentinel in 1.x branch.


flystar32:
LGTM

PR #230, opened 2019-01-02
use constructor injection instead of field injection

flystar32:
[Why field injection is evil](http://olivergierke.de/2013/11/why-field-injection-is-evil)

 -  unsafe code
 - more complicated to test

according to this, use constructor injection instead of field injection


pbting:
LGMT

PR #258, opened 2019-01-10
Enhance RocketMQ Binder error handle

fangjian0423:


spring-cloud-stream-binder-rocketmq/src/main/java/org/springframework/cloud/stream/binder/rocketmq/properties/RocketMQConsumerProperties.java
```diff
@@ -52,6 +52,40 @@
 
 	private Boolean enabled = true;
 
+	private ErrorProp error;
+
+	public static class ErrorProp {
```
xiaolongzuo: 建议Prop写全名。

fangjian0423: Maybe Error is ok? not ErrorProperties

xiaolongzuo: Yes.It looks better for use Error.

PR #260, opened 2019-01-10
reduce build time

zhendong590:
reduce build time

### DIFF
--- a/spring-cloud-alibaba-nacos-discovery/pom.xml
+++ b/spring-cloud-alibaba-nacos-discovery/pom.xml
@@ -82,5 +82,28 @@
         </dependency>
 
     </dependencies>
-
+    <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.codehaus.mojo</groupId>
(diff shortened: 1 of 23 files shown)


spring-cloud-alibaba-nacos-discovery/pom.xml
```diff
@@ -82,5 +82,28 @@
...
+            </plugins>
+        </pluginManagement>
+        <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>cobertura-maven-plugin</artifactId>
+                <configuration>
+                    <instrumentation>
+                        <excludes>
+                            <exclude>org/springframework/cloud/alibaba/nacos/**.*class</exclude>
```
flystar32: what's the purpose of excludes org/springframework/cloud/alibaba/nacos/**.*class here？ only this module need to exclude package here ?

这里的 exclude 是什么意思，为什么只在这个模块有？

zhendong590: my mistake ,i will fix it

PR #279, opened 2019-01-15
Refactor sentinel starter and add test case

fangjian0423:


spring-cloud-alibaba-sentinel/src/main/java/org/springframework/cloud/alibaba/sentinel/custom/SentinelBeanPostProcessor.java
```diff
@@ -60,10 +76,69 @@ public void postProcessMergedBeanDefinition(RootBeanDefinition beanDefinition,
...
+					"{} class property exists but {}"
+							+ " method property is not exists in bean[{}]",
+					type, type, beanName);
+			System.exit(-1);
+		}
+		else if (blockClass == void.class && !StringUtils.isEmpty(blockMethod)) {
+			logger.error(
+					"{} method property exists but {} class property is not exists in bean[{}]",
+					type, type, beanName);
+			System.exit(-1);
```
flystar32: Too violent to use System.exit(-1)，maybe just throw Exception is better

fangjian0423: yes, fixed now.

PR #305, opened 2019-01-22
feat: sentinel-feign support auto fallback 

pig4cloud:
Many times, the downgrade strategy is not useful, just want to output the log. But still have to write a Fallback class, and  inject into the Spring IOC。

So when @feignclient does not provide fallback, don't throw exceptions, output error logs by default, which is better, do you think?


fangjian0423:
I think provide a Default Fallback class is not a good idea because it could be created by users themselves.

Thank you for your contribution.

PR #384, opened 2019-02-24
use EnableDiscoveryClient

Harris2012:
add @EnableDiscoveryClient in example


flystar32:
@Harris2012   did you encountered any problem when there is no  `@EnableDiscoveryClient` ?

PR #446, opened 2019-03-18
commit pom change

a59871442:


a59871442:
更新

fangjian0423:
It is not necessary to add bom in example parent pom because it has been defined in spring-cloud-alibaba-dependencies module.

PR #502, opened 2019-03-28
dubbo examples cannot run because nacos port is miss.

BossYi:
dubbo examples cannot run while use nacos 1.0.0RC2 because nacos port is miss.
image
so i added the port

no

### Describe how you did it
add nacos port 8848 in bootstrap.yaml


fangjian0423:
Thank you for your contribution, we will review it soon.

PR #513, opened 2019-04-01
Add an usage for config listener. We can react to the change of config data.

MaoJianwei:
1. add an example that reacting to the change of config data using Listener.
2. update string format of output.
3. add default value for _user.age_, to avoid crash if Nacos server is not running.


MaoJianwei:
Could you please merge this commit? Thanks :)

PR #593, opened 2019-04-25
Nacos Cloud Service Registry add group name attribute

seas521:
I tested in a simple way ,please check it

### DIFF
--- a/spring-cloud-alibaba-nacos-discovery/src/main/java/org/springframework/cloud/alibaba/nacos/NacosDiscoveryProperties.java
+++ b/spring-cloud-alibaba-nacos-discovery/src/main/java/org/springframework/cloud/alibaba/nacos/NacosDiscoveryProperties.java
@@ -139,6 +139,12 @@ public class NacosDiscoveryProperties {
 	 */
 	private String secretKey;
 
+
+	/**
+	 * group name for nacos server.
+	 */
+	private String groupName = "DEFAULT_GROUP";
+
 	@Autowired
 	private InetUtils inetUtils;
 
@@ -337,6 +343,14 @@ public class NacosDiscoveryProperties {
 		this.watchDelay = watchDelay;
 	}
 
+	public String getGroupName() {
+		return groupName;
+	}
+
+	public void setGroupName(String groupName) {
+		this.groupName = groupName;
+	}
+
 	@Override
 	public String toString() {
 		return "NacosDiscoveryProperties{" + "serverAddr='" + serverAddr + '\''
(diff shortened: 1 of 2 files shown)


spring-cloud-alibaba-nacos-discovery/src/main/java/org/springframework/cloud/alibaba/nacos/NacosDiscoveryProperties.java
```diff
@@ -381,6 +396,11 @@ public void overrideFromEnv(Environment env) {
 			this.setEndpoint(
 					env.resolvePlaceholders("${spring.cloud.nacos.discovery.endpoint:}"));
 		}
+		if (StringUtils.isEmpty(this.getGroupName())) {
+			this.setEndpoint(
```
fangjian0423: It seems should invoke this.setGroupName(...)

seas521: My bad! pushed it again

flystar32:
@seas521  thank you for contribution to Spring-Cloud-Alibaba

I've asked @nkorange(https://github.com/nkorange), he said when you register a service with group, then you can only subscribe the service in the same group.

how do you think about this?

if this is the right way, we may use 

```
 List<Instance> selectInstances(String serviceName, String groupName, boolean healthy) throws NacosException;
```
in `NacosServerList.getServers()`

```
	private List<NacosServer> getServers() {
		try {
			List<Instance> instances = discoveryProperties.namingServiceInstance()
					.selectInstances(serviceId, true);
			return instancesToServerList(instances);
		}
		catch (Exception e) {
			throw new IllegalStateException(
					"Can not get service instances from nacos, serviceId=" + serviceId,
					e);
		}
	}
```

seas521:
> String

I think the group name is designed for setting apart from instances in same-name services，dont know if I am getting wrong,I just check the latest version of nacos-client1.0

fangjian0423:
I think group is a little bit repetitive like namespace.

If you register services with specified group, you should subscribe services with the same group.  

Now we could not subscribe all services in same namespace with all groups. If nacos support it, maybe this pull request can be merged.

chuntaojun:
> I think group is a little bit repetitive like namespace.
> 
> If you register services with specified group, you should subscribe services with the same group.
> 
> Now we could not subscribe all services in same namespace with all groups. If nacos support it, maybe this pull request can be merged.

if let user desc serviceName like this "serviceName@@group" and nacos can auto analyze group info，I think it can work

PR #606, opened 2019-04-26
Use YamlMapFactoryBean for yaml config loading to keep config-item order

losphantom:

The **spring-cloud-alibaba-nacos-config** use `YamlPropertiesFactoryBean` to load config in yaml format. That will load config item into a `Properties` which is unordered. There is some case we have to keep the config items order as we configured. 
e.g. When we configure `zuul routes` as follow:
```yaml
zuul.routes:
  service-1:
    path: /foo/**
  service-2:
    path: /bar/**
  service-x:
    path: /**
```
The order of the routes make sure when request a URI `/foo` will match `service-1` first.

NONE

### Describe how you did it

Use `YamlMapFactoryBean` instead, that load yaml config items into a `LinkedHashMap` as default.

### Describe how to verify it

Something like `zuul routes` config

### DIFF
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,1 +1,0 @@
--DaltSnapshotDeploymentRepository=repo.spring.io::default::https://repo.spring.io/libs-snapshot-local -P spring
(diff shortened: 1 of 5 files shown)


losphantom:
By the way, spring-boot use `YamlPropertySourceLoader` to load yaml format config which also keep config items order

PR #689, opened 2019-05-31
Dubbo graduated from Apache, remove the incubating word  in README.

lujian0571:

Dubbo graduated from Apache, remove the incubating word  in README.


fangjian0423:
It will redirect to dubbo if you visit incubator-dubbo.

Thanks for your contribution.

PR #774, opened 2019-07-23
Fix acm-local-example

Rivers-Shall:
Make the example app automatically refresh value from acm config server
by adding `@RefreshScope` annotation.

As indicated by the readme-zh.md of acm-local-example in spring-cloud-alibaba-examples.  The example app should automatically refresh the value `user.id` which is fetched from acm server.  The original code failed to do that and this PR fixes it.

NONE

### Describe how you did it

Adding `@RefreshScope` annotation.

### Describe how to verify it

Just run acm-local-example, change the configurations from acm server and see if the value in the app changes.

### Special notes for reviews

NONE


fangjian0423:
LGTM, thanks for your contribution.

PR #779, opened 2019-07-24
Fix the sample compile error

lovepoem:
Fix the sample compile error

 yes
### Describe how to verify it
ci success


fangjian0423:
thanks for your contribution

PR #780, opened 2019-07-24
fix #462 FileNotFoundException throws

### LINKED ISSUE #462: FileNotFoundException throws when rules is provided by local resources in classpath and executed in command line
Component: Sentinel

Description:
If the rules were configured in local file in classpath and executed the built jar in command line, FileNotFoundException throws, says,

class path resource [sentinel.json] cannot be resolved to absolute file path because it does not reside in the file system:jar:file:/D:/projects/a/b/c/blabal.jar!/BOOT-INF/classes!/sentinel.json

Investigation:
class: FileDataSourceProperties
method:preCheck

Rivers-Shall:
FileNotFoundException throws when rules is provided by local
resources in classpath and executed in command line

When we run the app in an IDE, everything is fine and the app is able to find
resource files.
When we pack our app into a jar with all the resource files in it and run it by 
`java -jar XXX.jar`, FileNotFoundException throws.
This PR makes sure that under both circumstances the app runs well.

Fixes #462 

### Describe how you did it

Modify `precheck` in `FileDataSourceProperties` to allow local files
in a jar.  Modify `FileRefreshableDataSourceFactoryBean` to return
different datasources according to file location---in the filesystem,
or in a jar.

### Describe how to verify it

**Rebuild** module `spring-cloud-alibaba-sentinel-datasource` and 
`sentinel-core-example` in `sentinel-examples` of `spring-cloud-alibaba-examples`.
Find the jar file under directory `sentinel-core-example/target` and run it by `java -jar XXX.jar`.
No exception is found and the rules are read from files in the jar.

### Special notes for reviews

The sentinel component offers the API which requests **name of the jar file** and **name of the file in  the jar**, which I found a little awkward(An API of `InputStream` will be far better and the internal implementation of current API is `InputStream`).  But anyway, I have to use string manipulation in `precheck` to extract **name of the jar file** and **name of the file in  the jar** from the URL.  Is there a better implementation?  Is there a better approach to handling exceptions?

### DIFF
--- a/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/FileDataSourceProperties.java
+++ b/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/FileDataSourceProperties.java
@@ -17,6 +17,8 @@
 package com.alibaba.cloud.sentinel.datasource.config;
 
 import java.io.IOException;
+import java.net.URL;
+import java.util.Arrays;
 
 import javax.validation.constraints.NotEmpty;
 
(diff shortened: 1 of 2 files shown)


zkzlx:
This pr was closed , you can reopen it after resolve conflicting files and the cla( Contributor License Agreement) .

PR #827, opened 2019-08-09
Fix nacos doc typo

zhaoyuguang:
Fix nacos doc typo. refer these issues:

#244


fangjian0423:
thanks for your contribution.

Issue #793, opened 2019-07-31
SentinelFeign not support FeignClient contextId！


Nitsuya:
**Which Component**
spring-cloud-alibaba-sentinel

**Describe the bug**
SentinelFeign not support contextId, can cause fallbackInstance not available, throw an exception

SentinelFeign中并未支持contextId，会导致fallbackInstance无法获取，抛出异常

**Expected behavior**
支持contextId

fangjian0423:
FeignClient add `contextId` attribute in Spring Cloud Greenwich.

contributors are welcome.

pigxcloud:
咋个回显出来。  contextId 只是给feignclient 搞了个 contexIdFeignClient 的名称 不会影响获取fallbackInstance啊

fangjian0423:
@pigxcloud 

https://github.com/alibaba/spring-cloud-alibaba/blob/master/spring-cloud-alibaba-sentinel/src/main/java/com/alibaba/cloud/sentinel/feign/SentinelFeign.java#L85

here. we should use `contextId`, not `name.

Nitsuya:
@pigxcloud name 和 contextId 配置成不一样的名字，再配置fallback， SentinelFeign L91->L110, 判空然后抛出异常

pigxcloud:
```
@FeignClient(contextId = "echoService", name = "service-provider", fallback = EchoServiceFallback.class)
public interface EchoService {

    /**
     * 调用服务提供方的输出接口
     *
     * @param str 用户输入
     * @return
     */
    @GetMapping(value = "/echo/{str}")
    String echo(@PathVariable("str") String str);
}
```

木有回显回来 好气

Nitsuya:
`
public class EchoServiceFallback implements EchoService {
    @Override
    public String echo(String str) {
        return "consumer-fallback-default-str" + str;
    }
}

public  class FeignConfiguration {
     @Bean
    public EchoServiceFallback echoServiceFallback() {
        return new EchoServiceFallback();
    }
}

@FeignClient(contextId = "echoService", name = "service-provider", fallback = EchoServiceFallback.class, configuration = FeignConfiguration.class)
public interface EchoService {
    @GetMapping(value = "/echo/{str}")
    String echo(@PathVariable("str") String str);
}
`

这样写就会报错, 
----
抱歉描述问题的时候没有贴出使用方式
排版不太会-,-.

pigxcloud:
> `
[...]

PR #838, opened 2019-08-13
fix spring cloud open feign in g version , sentinel-feign contextid bug

xiejiashuai:
fix spring cloud open feign in g version , sentinel-feign contextid bug

Fixes https://github.com/alibaba/spring-cloud-alibaba/issues/793

### DIFF
--- a/spring-cloud-alibaba-sentinel/src/main/java/com/alibaba/cloud/sentinel/feign/SentinelFeign.java
+++ b/spring-cloud-alibaba-sentinel/src/main/java/com/alibaba/cloud/sentinel/feign/SentinelFeign.java
@@ -26,6 +26,7 @@ import org.springframework.cloud.openfeign.FeignContext;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.util.ReflectionUtils;
+import org.springframework.util.StringUtils;
 
 import feign.Contract;
 import feign.Feign;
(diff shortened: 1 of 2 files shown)


spring-cloud-alibaba-sentinel/src/main/java/com/alibaba/cloud/sentinel/feign/SentinelFeign.java
```diff
@@ -82,20 +83,24 @@ public InvocationHandler create(Target target,
 							"fallback");
 					Class fallbackFactory = (Class) getFieldValue(feignClientFactoryBean,
 							"fallbackFactory");
-					String name = (String) getFieldValue(feignClientFactoryBean, "name");
+					String beanName = (String) getFieldValue(feignClientFactoryBean,
+							"contextId");
+					if (!StringUtils.hasText(beanName)) {
```
fangjian0423: I don't think we need to check the beanName from `contextId`, see https://github.com/spring-cloud/spring-cloud-openfeign/blob/master/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsRegistrar.java#L283.

if contextId is empty, it will get by `getName` method.

fangjian0423:
could you add some test cases about it?

spring-cloud-alibaba-sentinel/src/main/java/com/alibaba/cloud/sentinel/feign/SentinelFeign.java
xiejiashuai: I see this code, in g version ,   it will get by getName method and will get serviceId. As I do this , it compatible f version and g version

xiejiashuai:
> could you add some test cases about it?

Ok,i will add unit test about it

spring-cloud-alibaba-sentinel/src/main/java/com/alibaba/cloud/sentinel/feign/SentinelFeign.java
fangjian0423: ok, well done, and we need to modify the code in finchley & 1.x branch 😿

PR #850, opened 2019-08-15
some doc no error correction

panda-luo:
some doc no error correction

### Describe how you did it
fix some doc error.


panda-luo:
in MARKDOWN grammar: if has more than tow blank line, the serial number won't be in a row in view mode.

panda-luo:
so I delete two blank rows, and correct a doc no error.

fangjian0423:
thanks for your contribution.

PR #852, opened 2019-08-17
Optimize code and fix #841

### LINKED ISSUE #841: spring-cloud-dubbo-consumer-sample restTemplate  bug
When I comment out this annotation, the startup will report an error.
image

echooymxq:
* The `NacosPropertySourceLocator` check dataId file extension code is too redundant
*  decide file extension use `indexof()` instead of `endwith()` maybe cause a bug
* some spell check
NONE


fangjian0423:
i think all the strings with key word `nacos` shoude change to `Nacos`.

PR #865, opened 2019-08-19
Check nacos config health use configService

### LINKED ISSUE #855: NacosConfigHealthIndicator Always in UP state
**Which Component**
Nacos Config

**Describe the bug**
NacosConfigHealthIndicator Always in UP state

**To Reproduce**
1. project add spring-boot-starter-actuator
2. Stop Nacos service
3. Visit /actuator/health, Display results {"status":"UP"}

**Expected behavior**
{"status":"DOWN"}

**Screenshots**

The following code should add a judgement
```
builder.up().withDetail("dataIds", dataIds);
```
Change to
```
if( builder.build().getStatus() != Status.DOWN ) {
    builder.up().withDetail("dataIds", dataIds);
}
```

After modification, there is still a problem. Nacos also has cache configuration information locally. That is to say, when all Nacos services are deactivated, the health status can not be set to DOWN in time. Because configService. getConfig, the internal preference is to get the configuration from the local cache, so the configuration can be obtained, too. That is, health checks are invalid in this case.

If I want to do my personal sensory health check, then don't follow this logic. The easiest way is to read directly from the Nacos service every time. Reading exceptions can be set to DOWN at this time.

**Additional context**
spring-cloud-starter-alibaba-nacos-config.0.9.0.RELEASE

echooymxq:
The `NacosConfigHealthIndicator` has a bug with always in up state. 

Fixes #855

### DIFF
--- a/spring-cloud-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/endpoint/NacosConfigHealthIndicator.java
+++ b/spring-cloud-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/endpoint/NacosConfigHealthIndicator.java
@@ -16,16 +16,9 @@
 
 package com.alibaba.cloud.nacos.endpoint;
 
-import java.util.ArrayList;
-import java.util.List;
-
 import org.springframework.boot.actuate.health.AbstractHealthIndicator;
 import org.springframework.boot.actuate.health.Health;
-import org.springframework.util.StringUtils;
 
-import com.alibaba.cloud.nacos.NacosConfigProperties;
-import com.alibaba.cloud.nacos.NacosPropertySourceRepository;
-import com.alibaba.cloud.nacos.client.NacosPropertySource;
 import com.alibaba.nacos.api.config.ConfigService;
 
 /**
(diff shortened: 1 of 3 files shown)


spring-cloud-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/endpoint/NacosConfigHealthIndicator.java
```diff
@@ -33,15 +31,11 @@
...
 
 	private final ConfigService configService;
 
-	public NacosConfigHealthIndicator(NacosConfigProperties nacosConfigProperties,
-			ConfigService configService) {
-		this.nacosConfigProperties = nacosConfigProperties;
+	public NacosConfigHealthIndicator(ConfigService configService) {
 		this.configService = configService;
 
 		this.dataIds = new ArrayList<>();
```
fangjian0423: dataIds should be remove

echooymxq: I think The dataIds we can keep it, Form the dataid list, we can know ,what configuration have we loaded from nacos. private or ext or share.

fangjian0423: i prefer do it in `NacosConfigEndpoint`, not `NacosConfigHealthIndicator`. how do you think about it?

echooymxq: That is a better choice, i will remove it

fangjian0423: 👍

fangjian0423:
Test case runs failed.

[ERROR] Failures: 
[ERROR]   NacosConfigEndpointTests.contextLoads:93->checkoutAcmHealthIndicator:108 expected:<DOWN {}> but was:<UP {}>

PR #869, opened 2019-08-21
AsciiDoc headings should start with an equal sign(=)

zhaoyuguang:
AsciiDoc headings should start with an equal sign(=)


fangjian0423:
thanks for your contribution.

PR #871, opened 2019-08-21
add binder configuration properties test case

xiejiashuai:
Add some unit case for rocketmq binder


xiejiashuai:
For this pr , i am confused，for example i add some test case about binder properties ,but i don't know whether it is right, can give some help.

xiejiashuai:
For this pr , i am confused，for example i add some test case about binder properties ,but i don't know whether it is right, can give some help.

zkzlx:
please add test cases according to the latest version, this PR should be closed.

PR #876, opened 2019-08-23
Use Spring advanced API make design more clearer

zhaoyuguang:
We can use the more advanced Spring API provided to us to make the code clearer and more readable.


fangjian0423:
thanks for your contribution.

PR #880, opened 2019-08-26
 issue737 实现rocketmq的header的value的无感知序列化和反序列化

### LINKED ISSUE #737: spring-cloud-stream-binder-rocketmq Header value Codec Question
**Which Component**
spring-cloud-stream-binder-rocketmq

**Describe the bug**
After using the spring-cloud-stream-binder-rocketmq module, since the value of the header in the message in the rocketmq must be a string, the message header value is currently converted to a string using the toString method when the non-string is set.

I think this behavior is not appropriate.

Especially for numeric types such as int, long, etc., the receiver of the message is likely to think that the value should be the type of the number placed.

Moreover, the toString method of the normal object has no way to deserialize most of the time, because the toString method is not considered to be a data format, nor is it recommended to deserialize from the toString method to an object, so there is no ready-made tool class that can be used. 
Such an operation.

According to my understanding, Spring's design Since the Header value is Object, it should be the object that needs to be serialized and deserialized as both the header and the body. It should be serialized and deserialized in the same way.

It should be a set of HttpMessageConvert that is shared or a separate set of Http requests....

Now it’s a bit tangled, how to deal with it.

Because in theory the serialization format of Message can be any format including jdk serialization, json, xml, produf, the final format can be non-string.

This way there is no way to follow Spring's original design ideas.

使用spring-cloud-stream-binder-rocketmq模块后,由于rocketmq 中 message里的 header 的 value 必须是 string,所以 message header value设置非字符串时当前会直接使用toString方法转换成string.
我认为这样的行为不太妥当.
尤其是int,long之类的数字类型,消息接收方很可能会认为值应该就是放入的数字类型.
而且普通对象的toString方法大部分时候没有办法反序列化的,因为toString方法并不被认为是一个数据格式,也不提倡从toString方法反序列化为对象,因此也没有什么现成工具类可以用于这样的操作.

按照我的理解,Spring的设计既然Header值是Object,那其实应该是把Header和body都作为需要序列化和反序列化的对象的,应该是用同样的方式序列化和反序列化的.
应该是共用或者另起一套Http请求的那套HttpMessageConvert列表的.....

现在这样有点纠结,怎么处理好呢.
因为理论上Message的序列化格式可以是包括jdk序列化,json,xml,produf的任何格式,最终格式可以是非字符串的.
这样就没办法遵循Spring原来的设计想法了.

**Expected behavior**
Think of alternatives

1. ToString like this, with specific instructions and hints in the documentation and doc (still can't solve the problem that the object toString method is difficult to deserialize)

2. Error reporting, the caller must pass in a String by error, and force the caller to manually process

3. Fixed the final format of json or xml is the internal conversion of the string converter, which is converted in advance between the sender and the consumer, so that the user has no perception (hopefully this)

想到的可选方案
1.像现在这样ToString,在文档和doc进行特定说明和提示(还是无法解决对象toString方法难以反序列化的问题)
2.进行报错,通过报错约束调用方必须传入String,强制调用方人工处理
3.固定用json或者xml这种最终格式是字符串的转换器内部转换,在发送者和消费者中都提前转换,使用户无感知(希望是这种)

**Screenshots**

liudaomanbu:
Implementing senseless serialization and deserialization of value of rocketmq header
实现rocketmq的header的value的无感知序列化和反序列化

Fixes #737

### Describe how you did it
Designing an interface named RocketMQ Header Mapper is responsible for serialization and deserialization of header value, and adding an implementation class using jackson.

Then, in RocketMQListener BindingContainer and RocketMQMessageHandler, the pre-processing of sending and receiving messages is carried out to realize senseless serialization and deserialization.

At the same time, a consumer parameter spring. cloud. stream. rocketmq. bindings. < input-name >. consumer. trusted-packages is added to set the whitelist for deserialization.

设计一个名为RocketMQHeaderMapper的接口负责header的value的序列化和反序列化,并增加了一个使用jackson的实现类.
然后在RocketMQListenerBindingContainer和RocketMQMessageHandler中进行消息的发送和接受前置处理,以实现无感知序列化和反序列化.

同时增加了一个消费者参数spring.cloud.stream.rocketmq.bindings.<input-name>.consumer.trusted-packages用来设定反序列化时的白名单

### Describe how to verify it
A non-String type custom header is set when sending a message, and the message consumer verifies that the custom header of the message is deserialized to the type at the time of sending.

在发送消息时设置一个非String类型的自定义消息头,在消息消费者处验证该消息的自定义消息头是否反序列化为发送时的类型.


fangjian0423:
could you format code with eclipse code format? refer https://github.com/alibaba/spring-cloud-alibaba/wiki/%E5%A6%82%E4%BD%95%E8%B4%A1%E7%8C%AE%E4%BB%A3%E7%A0%81#%E9%85%8D%E7%BD%AE-spring-cloud-%E6%A0%87%E5%87%86%E7%9A%84%E4%BB%A3%E7%A0%81%E6%A0%BC%E5%BC%8F.

fangjian0423:
nice feature for RocketMQ Binder

Issue #483, opened 2019-03-26
Is the any way to walk around the exception when using custom response type on RestTemplate?


vipweihua:
Component: Sentinel

Description:
JsonParseException throws when we use custom response type on RestTemplate and traffic meets the threshold of the resource and fallback method invoked.

STR:
1. Define the API like http://localhost:8001/{id} which returns a User json.
2. Use RestTemplate on client-side to call the API and specify the response type as User.class
`
User user = restTemplate.getForEntity("http://localhost:8001/{id}", User.class, Collections.singletonMap("id", id)).getBody();
`
3. Exception  throws when calling the snippet code.

Investigation:
The blockHandler method defined on SentinelRestTemplate returns SentinelClientHttpResponse which has nothing containing User info.
Refer to the code:
class: HttpMessageConverterExtractor
method: public T extractData(ClientHttpResponse response) throws IOException

fangjian0423:
I think you could using `blockHandler` and `blockHandlerClass` of `@SentinelRestTemplate` to handle this:

```java
@SentinelRestTemplate(blockHandler = "handleException", blockHandlerClass = ExceptionUtil.class)
```

Then construct `User` in `ExceptionUtil#handleException`:

```java
public class ExceptionUtil {

	public static SentinelClientHttpResponse handleException(HttpRequest request,
			byte[] body, ClientHttpRequestExecution execution, BlockException ex) {
        ObjectMapper objectMapper = Utils.getObjectMapper();
        User user = new User(-1L, "block");
        try {
            return new SentinelClientHttpResponse(objectMapper.writeValueAsString(user));
        } catch (JsonProcessingException e) {
            // ignore
        }
        return new SentinelClientHttpResponse("RestTemplate request block by sentinel");
    }

}
```

It seems that we should define more `RestTemplate` when there is more than 1 POJO.

fangjian0423:
I think it can not walk around the exception because the convert action worked after interceptor.

zkzlx:
*my solution* :
1.Convert the degraded Response using the `MessageConvert` in `RestTemplate`;
   The `MessageConverter`  need to define a special `MediaType`, such as `degraded/degraded;charset=UTF-8` ; because its preconditions are distinguished by 'MediaType',at last,the 'MediaType' needs to be reset to the original value after the conversion is completed.
[...]

PR #906, opened 2019-09-03
Optimize the format of the returned data after the RestTemplate is degraded by Sentinel

zkzlx:

Optimize the format of the returned data after the RestTemplate is degraded by Sentinel

#483


fangjian0423:
It'a tricky way, i close it.

PR #920, opened 2019-09-07
fix(nacos): fix issue #859

chuntaojun:

fix issue #859 

### Describe how you did it

Because of the spring bean refresh mechanism, so I think to maintain a singleton object caching is more appropriate to hold the Service


fangjian0423:
I tried it, won't create new `ConfigService`, use same instance.

fangjian0423:
when RefreshEvent trigger, `ConfigService` will be created.

there are 2 solutions:

1. ConfigService as static field
2. Modify the internal logic of ConfigService

Issue #921, opened 2019-09-07
nacos 1.1.3 config data not exist


chuntaojun:
我们鼓励使用英文，如果不能直接使用，可以使用翻译软件，您仍旧可以保留中文原文。
We recommend using English. If you are non-native English speaker, you can use the translation software. 

**Which Component**

see nacos issue 1818：https://github.com/alibaba/nacos/issues/1818

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**
e.g.  MacOS  、Java8 、 Version 0.2.1.RELEASE

zkzlx:
Currently, it is rather rigid, and the subsequent [nacos ](https://github.com/alibaba/nacos)can support the optimization of suffix reading into dynamic

zkzlx:
I've done compatibility, and it's going to be dynamic

fangjian0423:
👍

PR #931, opened 2019-09-11
support reading nacos config with no suffix dataid &  enhance NacosConfigProperties

zkzlx:

Currently, it is rather rigid, I've done compatibility, and it's going to be dynamic
and the subsequent nacos can support the optimization of suffix reading into dynamic

#921

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ Examples：
 
 [Alibaba Cloud OSS Example](https://github.com/alibaba/spring-cloud-alibaba/blob/master/spring-cloud-alibaba-examples/oss-example/readme.md)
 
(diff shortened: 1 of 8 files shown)


fangjian0423:
The test cases run error because `SimpleReactiveDiscoveryClientAutoConfiguration` was added yesterday. 

Refer https://github.com/spring-cloud/spring-cloud-commons/blob/master/spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/simple/reactive/SimpleReactiveDiscoveryClientAutoConfiguration.java.

fangjian0423:
Need new test cases.

Issue #951, opened 2019-09-24
Unified version management using the dependencies module


ly641921791:
As titled.

Refer https://github.com/alibaba/spring-cloud-alibaba/pull/948#discussion_r326949488

Issue #911, opened 2019-09-04
BOM has dependency management for non direct dependencies


fangjian0423:
The current version has an opinion on RocketMQ Spring Boot and Dubbo Spring Boot.

PR #952, opened 2019-09-24
#951 Unified version management using the dependencies module

ly641921791:

#951 Unified version management using the dependencies module


fangjian0423:
hi, something is not correct.

we are on the progress about Spring Initializr, refer https://github.com/alibaba/spring-cloud-alibaba/issues/911.

ly641921791:
> hi, something is not correct.
> 
> we are on the progress about Spring Initializr, refer #911.

hi, I submitted a new code and didn't know if it was correct.

Issue #981, opened 2019-10-08
Support ReactiveDiscoveryClient with Nacos Discovery


fangjian0423:
Refer [ReactiveDiscoveryClient](https://github.com/spring-cloud/spring-cloud-commons/blob/master/spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/ReactiveDiscoveryClient.java).

echooymxq:
I want to finish it.

fangjian0423:
👍

PR #990, opened 2019-10-16
Support Nacos reactive service discovery

echooymxq:
 complete #981


fangjian0423:
hi, could you merge the newest commit? I fix the build error problem about circleci. thanks

fangjian0423:
Some problems:

1. import `NacosReactiveDiscoveryClientConfiguration ` in `NacosDiscoveryClientConfigServiceBootstrapConfiguration`. Spring Cloud Config need it
2. `NacosDiscoveryClientAutoConfiguration` need annotated with `ConditionalOnBlockingDiscoveryEnabled` Annotation
3. `NacosDiscoveryProperties` should be construct in `NacosReactiveDiscoveryClientConfiguration `

fangjian0423:
Need some AutoConfiguration Tests can avoid 2 & 3 problems.

goudai:
我想如果从nacos发起http请求的地方改为`Async Http Client`发起请求是不是更好一些，使用`Mono Flux`来包装和阻塞的HttpClient，这很丑陋。
类似下面的代码有一种掩耳盗铃的感觉.
```java
@Component
public class NacosReactiveDiscoveryClient implements ReactiveDiscoveryClient {

    @Autowired
    private DiscoveryClient discoveryClient;

    @Override
    public String description() {
        return discoveryClient.description();
    }

    @Override
    public Flux<ServiceInstance> getInstances(String serviceId) {
        return Flux.fromStream(discoveryClient.getInstances(serviceId).stream());
    }

    @Override
    public Flux<String> getServices() {
        return Flux.fromStream(discoveryClient.getServices().stream());

    }
}
```

echooymxq:
@goudai Sorry for delay reply. 
* The async http client, This is nacos scope.
* As your say, The Flux will blocking main thread on here, so there can be use a Scheduler, eg: using the bounded elastic scheduler.  contribution welcome.

echooymxq:
#1210

PR #991, opened 2019-10-16
fix circleci build problem

fangjian0423:


fangjian0423:
The [configuration](https://github.com/alibaba/spring-cloud-alibaba/blob/master/.circleci/config.yml#L13) of `resource_class` is large in circleci caused build fail, change to medium.

Issue #982, opened 2019-10-08
nacos config shared-dataids cont select group


khiker007:
我们鼓励使用英文，如果不能直接使用，可以使用翻译软件，您仍旧可以保留中文原文。
We recommend using English. If you are non-native English speaker, you can use the translation software. 

**Which Component**
Nacos Config

**Is your feature request related to a problem? Please describe.**
I want to get the configuration file from Nacos according to the environment, but Shared-data IDS can only be obtained from DEFAULT_GROUP
我想从nacos按照环境获取配置文件，但是shared-dataids只能从DEFAULT_GROUP获取

**Describe the solution you'd like**
shared-dataids: group:dubbo-consumer.yaml
可以指定group，不指定则默认使用DEFAULT_GROUP

zkzlx:
thanks for your suggestion, we will improve it later.

Issue #978, opened 2019-10-06
I hope that one of shared-dataids and ext-config can support profile recognition.


somowhere:
我们鼓励使用英文，如果不能直接使用，可以使用翻译软件，您仍旧可以保留中文原文。
We recommend using English. If you are non-native English speaker, you can use the translation software. 

**Which Component**
eg. Nacos Config

**Is your feature request related to a problem? Please describe.**
Microservices load the same public configuration can only be hard coded

**Describe the solution you'd like**
I hope that one of shared-dataids and ext-config can support profile recognition.

**Describe alternatives you've considered**
在 com.alibaba.cloud.nacos.NacosConfigProperties.Config 中增加是否启用profile属性，然后在NacosPropertySourceLocator.loadExtConfiguration 读取注入profile 
Add whether to enable the profile attribute in com.alibaba.cloud.nacos.NacosConfigProperties.Config and then read the injection profile in NacosPropertySourceLocator.loadExtConfiguration

zkzlx:
Thank you for your suggestion. I will support it as soon as possible.

zkzlx:
Configure load order : shared -> extensional -> application(default) ->application(suffix) -> application(profile).
There is therefore a priority for configuration items . 
The order of priorities is the reverse of the loading order.

fangjian0423:
Yes, different configurations have different order, refer https://github.com/alibaba/spring-cloud-alibaba/wiki/Nacos-config#%E9%85%8D%E7%BD%AE%E7%9A%84%E4%BC%98%E5%85%88%E7%BA%A7

Refer the order of Spring Boot Externalized Configuration: https://stackoverflow.com/a/25862357

PR #1008, opened 2019-10-21
[issue #982 & issue #978]enhance nacos config 

zkzlx:

enhance nacos config 

#982
#978


fangjian0423:
In my option, something need discuss:

1. configuration items like `extConfigs`, `refreshableDataIds` was added in `NacosConfigProperties` to replace `extConfig`, `refreshableDataids` and do some compatibility in post construct phase. I think new prefix can be added in new items and no need to do some compatibility, same prefix will cause confuse. Or don't add these new configuration items.
2. refactor conflicts with #1010, so i think no need to refactor in this PR.

zkzlx:
this is PR mainly to unify the configuration style and can enhance **shared-config**'s configuration freedom . it's not a core feature, so it's optional. i will close it later.

zkzlx:
Redesign the code structure

PR #1237, opened 2020-03-09
healthcheck switch config property

illlight:
add the health check exist in version < 1.5.1, 2.0.0 as an option.

we use the health check for k8s readnessprobe, find out that the new implement in version >= 1.5.1,2.0.0 make the status down because of unstable network, and the deployments restart.

applicaion works fine once get the configs from the nacos server, so i think there should have a switch to close the health check, and add back the old implement as an option with which the status will be up after the applicaion store the configuration in local cache.

### DIFF
--- a/spring-cloud-starter-alibaba/spring-cloud-starter-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/endpoint/NacosConfigEndpointAutoConfiguration.java
+++ b/spring-cloud-starter-alibaba/spring-cloud-starter-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/endpoint/NacosConfigEndpointAutoConfiguration.java
@@ -51,6 +51,8 @@ public class NacosConfigEndpointAutoConfiguration {
 	}
 
 	@Bean
+	@ConditionalOnMissingBean
+	@ConditionalOnProperty(name = "spring.cloud.nacos.config.health.enabled", matchIfMissing = true)
 	public NacosConfigHealthIndicator nacosConfigHealthIndicator() {
 		return new NacosConfigHealthIndicator(nacosConfigManager.getConfigService());
 	}
--- a/spring-cloud-starter-alibaba/spring-cloud-starter-alibaba-nacos-config/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/spring-cloud-starter-alibaba/spring-cloud-starter-alibaba-nacos-config/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -40,7 +40,6 @@
       "type": "java.util.List",
       "description": "a set of shared configurations .e.g: spring.cloud.nacos.config.shared-configs[0]=xxx ."
     },
-
     {
       "name": "spring.cloud.nacos.config.refreshable-dataids",
       "type": "java.lang.String",
@@ -55,6 +54,12 @@
       "name": "spring.cloud.nacos.config.extension-configs",
       "type": "java.util.List",
       "description": "a set of extensional configurations .e.g: spring.cloud.nacos.config.extension-configs[0]=xxx ."
+    },
+    {
+      "name": "spring.cloud.nacos.config.health.enabled",
+      "type": "java.lang.Boolean",
+      "defaultValue": true,
+      "description": "the switch for health check, it default enabled(true)."
     },
     {
       "name": "spring.cloud.nacos.config.refresh-enabled",


echooymxq:
Why not use the `spring boot actuator` for k8s probe？and check health with `/actuator/health` endpoint. Actually, Nacos's status cannot indicate the application status.

illlight:
@echooymxq this health check is one of the status of /actuator/health, the application status is down when it is down.

echooymxq:
@illlight I got it. I don't think we should distinguish remote or local caches check health. As long as the `NacosPropertySourceLocator` can locate `PropertySource`, the Nacos status should be `UP`.

echooymxq:
And i agree with you this point that we should add a property to decide whether to enable nacos health, eg:`spring.cloud.nacos.config.health.enabled`.

illlight:
@echooymxq That's it. And I also add the local config check just copy from the implement in version[ 1.5.0 ](https://github.com/alibaba/spring-cloud-alibaba/blob/v1.5.0.RELEASE/spring-cloud-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/endpoint/NacosConfigHealthIndicator.java) which checks the configuration could be read and not empty.

illlight:
The local config health check will not be enabled by default, or I can add it as an self defined health indicator.

illlight:
hope `spring.cloud.nacos.config.health.enabled`  add soon

echooymxq:
In my opinion, we just need revert the previous logic and add the `spring.cloud.nacos.config.health.enabled` property, `NacosLocalConfigHealthIndicator` is not necessary. @fangjian0423 what do you think?

mercyblitz:
> In my opinion, we just need revert the previous logic and add the `spring.cloud.nacos.config.health.enabled` property, `NacosLocalConfigHealthIndicator` is not necessary. @fangjian0423 what do you think?

I agreed with your opinion,  @illlight just remain the addion of `NacosConfigHealthIndicator`

illlight:
@mercyblitz done

spring-cloud-starter-alibaba/spring-cloud-starter-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/endpoint/NacosLocalConfigHealthIndicator.java
```diff
@@ -0,0 +1,82 @@
...
+			} catch (Exception e) {
+				builder.withDetail(String.format("dataId: '%s', group: '%s'",
+						dataId, nacosConfigProperties.getGroup()), "get config exception");
+			}
+		}
+		if (find) {
+			builder.up();
+		} else {
+			builder.down();
+		}
```
echooymxq: Use this part to replace current `NacosConfigHealthIndicator` health check code logic.

echooymxq:
@illlight I found this make things complicated.Just keep it will be ok.
```java
String status = configService.getServerStatus();
		// Set the status to Builder
		builder.status(status);
		switch (status) {
		case "UP":
			builder.up();
			break;
		case "DOWN":
			builder.down();
			break;
		default:
			builder.unknown();
			break;
		}
```

echooymxq:
And squash all your commits and adjust your commit message.

echooymxq:
Actually,I'm tangled here. If you followed my recommendation last night, then you only need to pay attention to `status`, and don't care about the details when unit testing.

echooymxq:
@illlight Can squash all your commits for one commit?

illlight:
@echooymxq ok

illlight:
@echooymxq I think both implementation are ok, what i really need is a switch, we can add a self-defined healch check to fit real scenarios.

echooymxq:
> @echooymxq I think both implementation are ok, what i really need is a switch, we can add a self-defined healch check to fit real scenarios.

If just so, we use with `@ConditionalOnMissingBean` and provide a default implement with configService.getServerStatus().

illlight:
@echooymxq That's fine, I hope it will be released in next version.

echooymxq:
@illlight  This can be done by modify your's pull request again. revert the `NacosConfigHealthIndicator` code, just use `configService.getServerStatus()` to check health, And add `@ConditionalOnMissingBean` support. Thanks.

illlight:
@echooymxq as your wish

echooymxq:
@illlight we also need support `spring.cloud.nacos.config.health.enabled`, And add some description to `additional-spring-configuration-metadata.json`.

Issue #1261, opened 2020-03-16
[Migration] Removing the commercial components from the code base


mercyblitz:
All Spring Boot or Spring Cloud starters should be removed from the code base, they will be migrated to a new project from Alibaba.

PR #1281, opened 2020-03-24
[Migration] Removing the commercial components from the code base

mercyblitz:
Issue : https://github.com/alibaba/spring-cloud-alibaba/issues/1261


fangjian0423:
README should also be modified to describe the components moving.

PR #1407, opened 2020-04-26
optimized code,A static field should be directly referred by its class name

XavierChengZW:
…s name

A static field or method should be directly referred by its class name instead of its corresponding object name.

Not

### Describe how you did it
A static field  be directly referred by its class name instead of its corresponding object name.

### Describe how to verify it
Not

### Special notes for reviews
Not


zkzlx:
the pr will be closed , you can reopen it after resolve conflicting files and the cla( Contributor License Agreement) . thks

PR #1463, opened 2020-05-19
Nacos support Common dataId application

hsoftxl:
Nacos support Common dataId application

支持公共配置dataId application
多环境下,自适应配置


zkzlx:
image

> My network can not comment on the submission record, so it has to leave a message in this way, sorry

Loading by a specified name is not recommended .
May affect existing usage, and it is better for the user to specify what configuration is required .

zkzlx:
[SeataFeignObjectWrapper](spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-seata/src/main/java/com/alibaba/cloud/seata/feign/SeataFeignObjectWrapper.java)
The modification here is the same as the other PR #1404 , you can solve the problem  leave a message in the #1404  PR, this PR will be closed, thank you.

PR #1487, opened 2020-05-27
fixes: alibaba#1478, support config Nacos ephemeral parameter

Gsealy:

support config nacos ephemeral in `application.properties`

Fixes: #1478


Gsealy:
same pr #1488

PR #1507, opened 2020-06-03
[#1506] fixed Properties file parsing problem

chuntaojun:

The configuration is resolved using springboot's own properties resolution code

https://github.com/alibaba/spring-cloud-alibaba/issues/1506


yuhuangbin:
refer: #1531

PR #1586, opened 2020-07-15
to avoid k8s readinessProbe remove the pod

qixiaobo:
For sentinel , we just no need remove the pod just because of the dashboard unreachable.
If sentinel dashboard crash , we will remove the pod just because indicator mark down. 
It's bad to do like this. 

https://github.com/alibaba/spring-cloud-alibaba/issues/842 

### Describe how you did it
We mark just UNKNOWN so the healthIndicator will return https status 200.

### Describe how to verify it
Killing dashboard , then will return 200 other then 503


qixiaobo:
Fix https://github.com/alibaba/spring-cloud-alibaba/issues/842

qixiaobo:
@fangjian0423 @mercyblitz Help~

fangjian0423 (approved):
LGTM

PR #1606, opened 2020-07-29
Add initial methods and implementation for gateway compatibility with dubbo.

dernDren161:
#### Initial implementation to receive requests from the gateway

Added new interface class and implemented it to receive different request parameters.


steverao:
We don't have relevant module now and close.

PR #1704, opened 2020-09-09
NacosWatch to update local metadata

### LINKED ISSUE #1701: Suggest sca to support dynamic metadata
want to change some metadatas via Nacos OPEN API or console UI. In Ribbon cache they will be updated by polling, but how to get the new changed metadatas in local service?

In my case, I use following codes to get local metadata, and metadata in Registration WILL never be updated

public Map<String, String> getMetadata() {
     return registration.getMetadata();
}

I think maybe it can work by DiscoveryClient to loop instances for getting the instance with new metadatas, but it is not so graceful. So suggest sca to support dynamic metadata

yuhuangbin:

Fixes #1701


flystar32 (approved):
LGM

mercyblitz (approved):

PR #1771, opened 2020-10-09
Timeout exception in feign first time calling

Flyfoxs:
When call feign API,  the first time always exception by timeout, not control by random seed.

It avoid time out to cause Exception.

### Describe how you did it
Give time buffer to feign client in application.properties


fangjian0423 (approved):
LGTM

PR #1939, opened 2021-01-21
fixed chinese garbled for nacos-config

zkzlx:


theonefx:
nacos (utf8) -> sca (convert to unicode) -> spring Resource

PR #2015, opened 2021-03-30
Some spaces are removed, which is more in line with the specification

tangyuewei:
Some spaces are removed, which is more in line with the specification
删除了一些空格，更符合规范

NONE


theonefx:
LGTM

PR #2039, opened 2021-04-15
fix issue #2038

### LINKED ISSUE #2038: Order is invalid where register ApplicationListener in DubboCloudRegistry#doSubscribe
**Which Component**
Dubbo 

**Describe the bug**
SCA 2.2.6-SNAPSHOT
SpringBoot 2.3.2.RELEASE
SpringCloud Hoxton.SR9

We should make the listener of the DubboMetadataService trigger before the listener of the normal service.
But in  DubboCloudRegistry, use @Order annotation on onApplicationEvent method, it's invalid. 

So, the listeners has same order. listener trigger depends on the order of registration.

If we start the consumer first and then the provider, then the listener of the normal service trigger before DubboMetadataService. When nacos offline provider, the service cannot be called after restart. Because the DubboMetadataService has not been update  URL when update normal service URL, request DubboMetadataService failed and the update of the normal service URL failed.

**To Reproduce**
1. start consumer
2. start provider
3. call provider, it's ok
4. nacos offline provider
5. restart provider, call provider, No provider Error

Jisfeng:

We should make the listener of the DubboMetadataService trigger before the listener of the normal service.
But in DubboCloudRegistry, use @order annotation on onApplicationEvent method, it's invalid.

fix #2038 

### Describe how you did it

匿名类实现接口，使用GenericApplicationListener替代ApplicationListener，重写getOrder方法

### DIFF
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-dubbo/src/main/java/com/alibaba/cloud/dubbo/registry/DubboCloudRegistry.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-dubbo/src/main/java/com/alibaba/cloud/dubbo/registry/DubboCloudRegistry.java
@@ -42,10 +42,12 @@ import org.slf4j.LoggerFactory;
 
 import org.springframework.cloud.client.ServiceInstance;
 import org.springframework.cloud.client.discovery.DiscoveryClient;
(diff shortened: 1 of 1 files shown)


Jisfeng:
> this issue has been resolved by #2071

ok

Issue #2079, opened 2021-05-18
[Bug] Dubbo Spring Cloud cause "o provider available from registry"


fangjian0423:
Now [Dubbo Spring Cloud](https://github.com/alibaba/spring-cloud-alibaba/tree/master/spring-cloud-alibaba-starters/spring-cloud-starter-dubbo) will cause "org.apache.dubbo.rpc.RpcException: No provider available from registry ..." exception in Kubernetes Cluster when Application provider pod restart.

The problem still exists after some code fix.

This issue want to tell community the bug. Be vigilant if you want to use Dubbo Spring Cloud in production environment。

We hope some guys could help to reproduce and fix it, Thanks!

Refer https://github.com/alibaba/spring-cloud-alibaba/issues/1805.

PR #2101, opened 2021-06-01
use revision to optimize dubbo registry

theonefx:
- attach revision info when registry
- group by revision when subscription
- use sequential logic to process to process metadataservice and generalservice

This PR will make the service subscription logic more stable and efficient


ppzzyy11:
👍

yumin5723:
Will the next version include a fix for this bug ? [https://github.com/alibaba/spring-cloud-alibaba/issues/2079](url)

theonefx:
> Will the next version include a fix for this bug ? [https://github.com/alibaba/spring-cloud-alibaba/issues/2079](url)

This PR is designed to fix this problem, and we will release them in next version. you can try it with version 2.2.6-bugfix4-SNAPSHOT

Issue #2094, opened 2021-05-26
Sentinel对Feign限制粒度以及扩展问题


A11enHuang:
**Which Component**
 Sentinel

**Is your feature request related to a problem? Please describe.**
Sentinel对Feign的支持中，每一个rest接口都会当作一个资源进行限流处理，可是Sentinel对资源的数量会有限制。目前的版本(1.8)中未提供对此资源限制的调整。如果项目中本身要限制的资源太多，那么就会造成直接忽略对某些资源的限制。并且SentinelFeign是final类，SentinelInvocationHandler的构造方法也仅限于本包中。因此如果我想要进行扩展，是没法做到的。

In Sentinel's support for Feign, each rest interface will be treated as a resource for current limiting, but Sentinel will limit the number of resources. The current version (1.8) does not provide adjustments to this resource limit. If there are too many resources to be restricted in the project itself, it will cause the direct ignorance of the restrictions on certain resources. And SentinelFeign is a final class, the construction method of SentinelInvocationHandler is also limited to this package. So if I want to expand, I can't do it.

**Describe the solution you'd like**
能不能开放对资源限制粒度的配置。例如我针对某一个service进行限制，或者同一个service中的某一个domain进行限制，最小的粒度才是针对rest接口的限制。
在SentinelFeign中提供重载方法，让我能对构建过程进行扩展，或者能够提供装饰器，让我对构建进行扩展。

Can you open up the configuration of resource restriction granularity. For example, if I restrict a certain service, or restrict a certain domain in the same service, the smallest granularity is the restriction for the rest interface.
Provide overloaded methods in SentinelFeign, so that I can extend the build process, or can provide a decorator, let me extend the build.

yuhuangbin:
欢迎贡献 PR

Cuihongsen:
我这边需要一个能自定义资源名的方式，这个Feature 似乎有用

DanielLiu1123:
see #2342

PR #2106, opened 2021-06-04
feat: sentinel对feign的资源粒度的支持。

A11enHuang:
feat: sentinel对feign的资源粒度的支持

当前的sentinel处理feign时，无法控制资源的粒度。每一个rest api都当作一个粒度。这并不合理。当一个服务出现问题时，该服务的其他接口也很有可能也无法访问，所以按照服务为粒度进行资源限制是一个合理的需求。并且当前版本的sentinel最多只能对2000个资源进行限制，如果每一个rest api都当作一个资源来限制的话，有可能造成部分资源无法被sentinel处理的问题。

When the current sentinel processes feign, it cannot control the granularity of resources. Each rest api is treated as a granularity. This is not reasonable. When a service has a problem, other interfaces of the service may also be inaccessible, so it is a reasonable requirement to restrict resources according to the granularity of the service. And the current version of sentinel can only limit up to 2000 resources. If each rest api is treated as a resource to limit, it may cause the problem that some resources cannot be processed by sentinel.

Feature #2094

### Describe how you did it
在SentinelInvocationHandler生成resourceName时，将会查找生成策略，根据不同的生成策略来进行不同的处理。全局默认的生成策略为rest api.可以在配置文件中(spring.cloud.sentinel.feign.resource-handler.default-handler)配置默认的生成策略。
当然，为了让粒度控制变得更加灵活，我也自定义了一个注解: @ResourceHandler 他可以作用在类上，也可以作用在方法上，该注解能够指定要使用的策略。其中粒度设置优先级的关系是 方法 > 类 > 全局配置。用户也可以根据自己的实际情况，定义自己的粒度策略。

When the SentinelInvocationHandler generates the resourceName, it will look up the generation strategy and perform different processing according to different generation strategies. The global default generation strategy is rest api. The default generation strategy can be configured in the configuration file (spring.cloud.sentinel.feign.resource-handler.default-handler).
Of course, in order to make the granular control more flexible, I also customized an annotation: @ResourceHandler can act on the class or on the method. The annotation can specify the strategy to be used. The relationship of granularity setting priority is Method> Class> Global Configuration. Users can also define their own granularity strategy according to their actual situation.

### Describe how to verify it
com.alibaba.cloud.sentinel.SentinelFeignTests#testFeignServiceInstanceResourceHandler()
可以在接口中添加注解，调整resource策略。

com.alibaba.cloud.sentinel.feign.SentinelInvocationHandler第115行，断点可以看到生成的resourceName。

### DIFF
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-sentinel/src/main/java/com/alibaba/cloud/sentinel/feign/SentinelFeign.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-sentinel/src/main/java/com/alibaba/cloud/sentinel/feign/SentinelFeign.java
@@ -16,24 +16,15 @@
 
 package com.alibaba.cloud.sentinel.feign;
 
-import java.lang.reflect.Field;
-import java.lang.reflect.InvocationHandler;
-import java.lang.reflect.Method;
-import java.util.Map;
-
 import feign.Contract;
 import feign.Feign;
 import feign.InvocationHandlerFactory;
-import feign.Target;
-import feign.hystrix.FallbackFactory;
 import feign.hystrix.HystrixFeign;
 
 import org.springframework.beans.BeansException;
-import org.springframework.cloud.openfeign.FeignContext;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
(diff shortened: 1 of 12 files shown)


DanielLiu1123:
Supported. see #2342

Issue #2300, opened 2021-11-17
enable legacy bootstrap way


zchazc:
**Which Component**
spring-cloud-starter-alibaba-nacos-config

**Describe the bug**
因新版`spring-cloud`弃用`bootstrap.properties`导致的传统配置方式不生效问题。
The legacy  configuration method does not take effect due to the new `spring-cloud` abandonment of `bootstrap.properties`.

**To Reproduce**
#2048 

**Expected behavior**
- Main:
spring-cloud [reference](https://docs.spring.io/spring-cloud-config/docs/3.0.4/reference/html/#config-first-bootstrap)
spring-cloud-starter-alibaba-nacos-config [reference](https://github.com/alibaba/spring-cloud-alibaba/blob/a064331b88e59e0ae48fe512a00d3c2da8989928/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-config/pom.xml#L36-L40)
- Others:
spring-cloud-starter-bootstrap [only one file](https://github.com/spring-cloud/spring-cloud-commons/blob/main/spring-cloud-starter-bootstrap/src/main/java/org/springframework/cloud/bootstrap/marker/Marker.java),
mismatch with [example](https://github.com/alibaba/spring-cloud-alibaba/blob/master/spring-cloud-alibaba-examples/nacos-example/nacos-config-example/readme.md#connect-to-nacos-config)

**Additional context**
能提个 PR 把 optional 删了吗?
Can I create a PR to delete `optional`?

steverao:
Sure, welcome to create a PR to fix the problem!

zchazc:
> Sure, welcome to create a PR to fix the problem!

Okey, I done~😊 #2301

PR #2301, opened 2021-11-18
fixup: issue#2300

zchazc:
Use the legacy bootstrap way in newer spring-cloud version

#2300 #2048 

All about this pr in issue #2300


steverao:
LGTM.

PR #2376, opened 2022-01-27
Modify the customize() method to facilitate subclass override

galaxy-sea:

Modify the customize() method to facilitate subclass override


galaxy-sea:
``private static void customize()`` 在Spring托管的bean中看到``static``方法 真的会觉得很突兀

PR #2378, opened 2022-01-27
dependency optimization

galaxy-sea:
2. delete invalid groupId
3. add XML declaration

1. delete duplicate dependencies
2. delete invalid groupId
3. add XML declaration


DanielLiu1123:
Unnecessary change.

PR #2403, opened 2022-02-21
Upgrade nacos 1.4.2

DanielLiu1123:
Tests passed.
Upgrade plugins version, keep same as Spring boot.
Introducing [Testcontainers](https://github.com/testcontainers/testcontainers-java/) to do integration tests.

NOTE: nacos 1.4.3 replaces the verification method, can no longer use `localhost:8848`, use ` 127.0.0.1:8848` to replace it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -93,15 +93,16 @@
         <rocketmq.version>4.6.1</rocketmq.version>
 
         <!-- Maven Plugin Versions -->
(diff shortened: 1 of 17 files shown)


steverao:
Due to some problems with the authentication method of Nacos 1.4.3, we decided to use Nacos 1.4.2!

PR #2482, opened 2022-03-25
add spring-cloud-alibaba-schedulerx starter and examples

HuangXiaomeng:


steverao:
This is not suitable for putting in spring cloud alibaba!

PR #2526, opened 2022-05-05
Modified the log level of RocketMQProducerMessageHandler.

perfectism13:
image

The log level here should be debug.


steverao:
LGTM.

PR #2579, opened 2022-05-30
[ISSUES #2577] Update Seata to 1.5.1 

### LINKED ISSUE #2577: Update Seata to 1.5.1 in Spring Cloud Alibaba 2.2.x branch
我们鼓励使用英文，如果不能直接使用，可以使用翻译软件，您仍旧可以保留中文原文。
We recommend using English. If you are non-native English speaker, you can use the translation software.

Which Component
eg. Seata,

Is your feature request related to a problem? Please describe.
We are prepared to release a new version of Spring Cloud Alibaba in 2.2.x branch, we are plan to update some components.

Describe the solution you'd like
Update Seata to 1.5.1 in Spring Cloud Alibaba 2.2.x branch, and text the reliability.

zhangqian9158:

see the issue #2577
add redisLocker's lua mode
support the mixed use of different storages of locks and sessions
add a Executor for INSERT ON DUPLICATE KEY UPDATE
provide an api to share tcc phase-1's params to phase-2
support configuring the order of the TM and TCC interceptor
support configuring scan target for GlobalTransactionScanner
support redis distributed lock to prevent multi TC competition
TCC mode support idempotent and anti hanging
support server start with springboot and config with application.yaml
support APM with SkyWalking
TCC mode supports customized parameters list of the method in phase two
TCC mode's try method supports passing BusinessActionContext implicitly
support edas-hsf RPC framework
contributing md support chinese.
support GlobalTransactionInterceptor expression
support the registry center network preferences
support get configuration from environment
support SPI unload
support kotlin coroutine
support brpc-java RPC framework
init the console basic code
query global session in the file mode
query global session and global lock in the redis mode
get global lock in the file mode
Realize configuration center upload configuration interactive script (nacos,etcd3)
Realize configuration center upload configuration interactive script (apollo,consul,zk)
realize the interface of console: get global session and global lock in the db mode
console front-end page implementation
implementation of DefaultAuthSigner
make seata-bom be the real Bill-Of-Material
add db realization for distribute lock
registry add heartbeat
support zstd compressor
Saga support auto configuration in the spring boot project
 
Fixes: #2577 
fix tcc phase two response timeout exception
fix NPE and wrong cluster name of Apollo
fix the problem in the findTargetClass method
fix typo of interval
fix consul not found tc cluster
fix mariadb unable to create XA connection
fix the problem that store mode does not take effect
fix that LocalThread is not cleared when the Saga transaction ends
fix the Server can't find redis-host property
fix StringUtils StackOverflowError
fix TC SkyWalking topo calling node not gather
fix ReflectionUtil throw unexpected exception
fix postgresql multi schema throw not found channel exception
fix getConfig with different default value return the first
fix LocalDataTime type in FastjsonUndoLogParser can't be rollback
fix seataio/seata-server servlet-api conflict
fix the wrong path and filename when dump the jvm memory for analysis
fix NPE cause by future timeout
fix register branch and release lock failed when the size of rows that modified is greater than 1000 in oracle
fix the problem that nacos-config.py will not skip blank options. fix bug that split options may cause content loss
fix the problem that nacos not found user when password has special characters
fix the NPE of jedis multi.exec
fix can not get properties of distributed-lock-table in springboot
fix potential database resource leak
fix the problem that the xid is not cleared in some scenes of dubbo
fix RM did not clear XID after the local transaction threw an exception
fix ApplicationContext already closed problem when Seata server using ShutdownHook to destroy
fix prevents XA mode resource suspension
fix deadlock problems during project construction
fix the logback can't load the RPC_PORT
fix correct built-in properties for redis registry
fix StringUtils.toString(obj) throw ClassCastException when the obj is primitive data array
fix xa mode originalConnection has been closed, cause PhaseTwo fail to execute
fix the problem of accidentally releasing the global lock
fix delete undo log connection already closed
fix the kafka-appender.xml and logstash-appender.xml
fix code for "sessionMode" not execute problem
fix some problems with zstd compressor and add the version of the kotlin-maven-plugin
fix could not rollback when insert field list is empty
update executor store the actually modified columns but not only the columns in set condition
fix seata-test module UT not work
fix the problem that mysql's Blob/Clob/NClob data type cannot be deserialized
fix the problem that other ORMs may not be able to obtain the auto-incrementing primary key value
fix data remanence problems in lock and branch under specific circumstances.
fix the TableMetaCache parsing problem with the same table under multiple Postgresql schemas
fix inability to build Executor when using mariadb driver
fix mysql-loadbalance resource id error
fix the problem that failed to obtain the self increment ID of MySQL database through "select last_insert_id"
fix dirty write check exception that may occur when using ONLY_CARE_UPDATE_COLUMNS configuration
fix resource suspension in xa mode caused by choose other ip as channel alternative
fix the invalid environment variable in container env
fix the problem that pipelined resources are not closed in redis mode and add branchSession judge branchSessions is not null
fix the problem that GlobalSession could not be deleted normally in the case of delayed deletion in the file mode of the develop branch
fix the inability to get some remote configurations
fix the change log of 'service.disableGlobalTransaction' config
fix redis mode page npe and optimize get globalSession on average
fix the failure to obtain before image and after image on oracle and pgsql of the develop branch
fix Mysql multi-bit Bit type field rollback error
fix the failure to update cluster list dynamically when use eureka of the develop branch
fix FileSessionManagerTest fail
fix allSessions/findGlobalSessions may return null and cause npe
fix fastjson serialization of time data types
fix prepareUndoLogAll of MySQLInsertOrUpdateExecutor
fix PK constraint name isn't the same as the unique index name which is belong to PK
fix saga complex parameter deserialization problem
fix rpc tm request timeout
fix some problem of the sql parser

### DIFF
--- a/spring-cloud-alibaba-dependencies/pom.xml
+++ b/spring-cloud-alibaba-dependencies/pom.xml
@@ -20,7 +20,7 @@
     <properties>
         <revision>2.2.8-SNAPSHOT</revision>
         <sentinel.version>1.8.4</sentinel.version>
-        <seata.version>1.4.2</seata.version>
+        <seata.version>1.5.1</seata.version>
         <nacos.client.version>2.1.0</nacos.client.version>
         <nacos.config.version>0.8.0</nacos.config.version>
         <spring.context.support.version>1.0.11</spring.context.support.version>


steverao:
Please sign the Contributor License Agreement! By the way, can you provider some examples or add some documents about the new features in  seata 1.5.1.

Issue #2593, opened 2022-06-13
Support register ipv6 address to Nacos


ralf0131:
我们鼓励使用英文，如果不能直接使用，可以使用翻译软件，您仍旧可以保留中文原文。
We recommend using English. If you are non-native English speaker, you can use the translation software. 

**Which Component**
Nacos

**Is your feature request related to a problem? Please describe.**
Support register ipv6 address to Nacos
When I have both ipv4 and ipv6 address, SCA seems to register ipv4 address to Nacos, we provider an option for user to register either ipv4 , ipv6 or both.

cheese8:
I can take this,  assign me if in convenient.

HHoflittlefish777:
i can do it

PR #2604, opened 2022-06-16
add preferredIPversion to resolve ipv4/ipv6  address or both

cheese8:
add preferredIPversion to register for your service instance with ipv4, ipv6 or both.

Ref #2593

### Describe how you did it
As coded, get ipv4 while v4 preferred, get ipv6 while v6 preferred, return ip with `IPV4,IPV6` if both preferred and register for your service instance twice later.

And how about to enhance org.springframework.cloud.commons.util.InetUtils with ipv6?  @ralf0131 

### Describe how to verify it
not verified yet.

### Special notes for reviews
NONE

### DIFF
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/main/java/com/alibaba/cloud/nacos/NacosDiscoveryProperties.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/main/java/com/alibaba/cloud/nacos/NacosDiscoveryProperties.java
@@ -16,19 +16,6 @@
 
 package com.alibaba.cloud.nacos;
 
-import java.net.Inet4Address;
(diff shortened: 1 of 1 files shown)


steverao:
You just think about resolving IPv6 address, you don't care RPC invocation. When you get IPv6 address from registry, if use IPv6 to create a request just like IPv4, maybe it will get error.

steverao:
You just think about resolving IPv6 address, you don't care RPC invocation. When you get IPv6 address from registry, if use IPv6 to create a request just like IPv4, maybe it will get error.

cheese8:
> You just think about resolving IPv6 address, you don't care RPC invocation. When you get IPv6 address from registry, if use IPv6 to create a request just like IPv4, maybe it will get error.

ok, thanks for your  reminder, I will test it later. This pr is planed to resolving the IPv6 address, I changed the pr title.

steverao:
#2628

PR #2784, opened 2022-09-16
Fix issue 2696: remove some invalid test cases. Because no Nacos server now.

### LINKED ISSUE #2696: Remove some invalid test cases
In branch of 2.2.x, there are some test cases, they run depend on Nacos server. But we don't Nacos server now, it will cause run failed in process of the CI. Then, we need to remove them.

pandaapo:
Remove some invalid test cases which depend on Nacos server. Because no Nacos server now.

Fixes #2696

### Describe how to verify it
run `mvnw test`


steverao:
It's not the problem you did, I solved it in #2871

PR #2785, opened 2022-09-16
fix: spring cloud bus default environment initialization error

pumbf:
In 2021.x branch, spring-cloud-starter-stream-rocketmq has been refactored.However, the bus is not adapted to the new properties. Lead consumption group mode is CLUSTERING but not BROADCASTING.

NONE
### Describe how you did it
change the spring-cloud-starter-stream-rocketmq EnvironmentPostProcessor

### Describe how to verify it
Open two customers, see if it is possible to receive remoteEvent


steverao:
Please explain more about why we need to modify to BROADCASTING.

pumbf:
Spring Cloud Bus links nodes of a distributed system with a lightweight message broker. This can then be used to broadcast state changes (e.g. configuration changes) or other management instructions.  that;s mean all node need to receive message. At the same time, the automatic configuration of the old version also set the mode to BROADCASTING.
eg:
```java
String broadcastingPropertyName = createRocketMQPropertyName(INPUT,"broadcasting");
source.put(broadcastingPropertyName, "true");
```
So the main problem of spring cloud bus is that the new version does not synchronize the stream configuration changes, resulting in the inability to set the correct mode

pumbf:
@steverao Can this be explained?

laizuan:
2022.x 版本也存在这个问题，新版本中 broadcasting 属性已经不生效了，需要配置messageModel

image

PR #2806, opened 2022-09-25
When ipType is IPv6 but there's no IPv6 found, find IPv4(Fixes gh-2802)

### LINKED ISSUE #2802: Register a invalid IPv6 address
**Which Component**
eg. Nacos Discovery

**Describe the bug**
If I use `spring.cloud.inetutils.preferred-networks=00` to filter the invalid IPv6 address, it will return [127.0.0.1], that is a invalid IPv6 address.

image

If I call the service by [127.0.0.1], it will print error information like below:
```
java.net.URISyntaxException: Malformed IPv6 address at index 8: http://[127.0.0.1]:20003/c
	at java.net.URI$Parser.fail(URI.java:2845) ~[na:1.8.0_291]
	at java.net.URI$Parser.parseIPv6Reference(URI.java:3466) ~[na:1.8.0_291]
	at java.net.URI$Parser.parseServer(URI.java:3216) ~[na:1.8.0_291]
	at java.net.URI$Parser.parseAuthority(URI.java:3152) ~[na:1.8.0_291]
	at java.net.URI$Parser.parseHierarchical(URI.java:3094) ~[na:1.8.0_291]
	at java.net.URI$Parser.parse(URI.java:3050) ~[na:1.8.0_291]
	at java.net.URI.<init>(URI.java:673) ~[na:1.8.0_291]
```
So, I think if we can't find a valid IPv6 address, we should register a valid IPv4.

pandaapo:
When `spring.cloud.nacos.discovery.ip-type` is IPv6 but no IPv6 can be found out, SCA will automatically find IPv4. Thus a available service address is more likely to be provided. 

Fixes #2802

### Describe how to verify it
Use a machine which doesn't support IPv6 to register. Or in application, set `spring.cloud.inetutils.preferred-networks` to exclude any IPv6.

### DIFF
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/main/java/com/alibaba/cloud/nacos/NacosDiscoveryProperties.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/main/java/com/alibaba/cloud/nacos/NacosDiscoveryProperties.java
@@ -269,7 +269,7 @@ public class NacosDiscoveryProperties {
 				else if ("IPv6".equalsIgnoreCase(ipType)) {
 					ip = inetIPv6Util.findIPv6Address();
 					if (StringUtils.isEmpty(ip)) {
-						log.error("There is no available IPv6 found. SCA will automatically find IPv4.");
+						log.warn("There is no available IPv6 found. Spring Cloud Alibaba will automatically find IPv4.");
 						ip = inetUtils.findFirstNonLoopbackHostInfo().getIpAddress();
 					}
 				}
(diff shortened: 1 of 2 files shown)


spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/main/java/com/alibaba/cloud/nacos/NacosDiscoveryProperties.java
```diff
@@ -266,6 +268,10 @@ public void init() throws Exception {
 				}
 				else if ("IPv6".equalsIgnoreCase(ipType)) {
 					ip = inetIPv6Util.findIPv6Address();
+					if (StringUtils.isEmpty(ip)) {
+						log.error("There is no available IPv6 found. SCA will automatically find IPv4.");
```
steverao: I think use `warn` is more appropriate. by the way, use Spring Cloud Alibaba instead of SCA.

steverao:
By the way, please create a same PR to branch of 2.2.x.

pandaapo:
> By the way, please create a same PR to branch of 2.2.x.

OK

steverao:
LGTM.

PR #2884, opened 2022-11-02
registry instance-id and set default value

horizoen:

- In some scenarios, the service instance id is required as a registration or filter condition，and the version(tag: 2021.1) can not registry service instance-id and  get service instance-id is null

### Describe how you did it

- Set default values
- Add features related to instance ids, such as registration and discovery


steverao:
Why do you need to register a new field called instanceId? If you need it why don't construct it by instance lists in consumer?

galaxy-sea:
instanceId应该由nacos service注册成功后返回，不该Spring Cloud Alibaba这边自定义一个instanceId。
个人觉得nacos社区应该对 ``nacos/v1/ns/instance``返回值进行一个重构，放在返回值中增加instanceId

steverao:
We solved the problem in new release. In fact, generating the instanceId always do in Nacos Server. Now, you can get instanceId in Spring Cloud Alibaba if relevant version of Nacos Server support generating it. You can update to Spring Cloud Alibaba 2021.0.1.0.

PR #2934, opened 2022-11-29
[ISSUE #2906] Fix same dataId with different group and namespace can'…

### LINKED ISSUE #2906: Support same dataId with different group and namespace can't override in local
### Background
Spring Boot 2.4 introduced a new way to import configuration data via the spring.config.import property. 
But in local environment, if we add following properties in application.yml:
```
spring:
  config:
    import:
      - optional:nacos:test2.yml?group=DEFAULT_GROUP&refreshEnabled=true
      - optional:nacos:test2.yml?group=test&refreshEnabled=true
```  
The one will override another. For specific reason, that is because we create propertySource name by dataId, we should construct the name by `dataId@group@namespace`.

### How to do
You can refer to `spring-cloud-starter-alibaba-nacos-config` module in branch of 2021.x and understand how we pull config from config center by spring.config.import.

ruansheng8:
Fix same dataId with different group and namespace can't override in local

Fixes #2906


steverao (approved):

steverao:
LGTM, thanks @ruansheng8

PR #2945, opened 2022-12-02
 fix some invalid urls in appactive module

steverao:
Fix some invalid urls stored in other cloud server in appactive module


theonefx:
LGTM

theonefx (approved):

PR #2949, opened 2022-12-05
Reuse rocketmq producer in 2021.x

LeBW:
Reuse rocketmq producer in 2021.x


steverao (approved):
LGTM, thanks @LeBW

PR #2950, opened 2022-12-05
Change default group name to "anonymous"

LeBW:
As there has been a constant in `RocktetMQConst`, there is no need to add a constant in `RocketMQProduceFactory` .


steverao (approved):
LGTM, thanks @LeBW

PR #2970, opened 2022-12-13
Change the package name of javax in the example to Jakarta.

Phixsura:
Change some package name of javax in the example to Jakarta.


steverao (approved):
LTGM, thanks!

PR #2987, opened 2022-12-20
SCA examples native support

ruansheng8:
1. add graalvm and spring boot build plugin to examples 
2. update seata version to 1.6.1-SNAPSHOT

### DIFF
--- a/spring-cloud-alibaba-examples/pom.xml
+++ b/spring-cloud-alibaba-examples/pom.xml
@@ -56,16 +56,14 @@
 		<module>integrated-example/integrated-frontend</module>
 	</modules>
 
-    <properties>
-        <seata.version>1.6.1-SNAPSHOT</seata.version>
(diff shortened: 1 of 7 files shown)


spring-cloud-alibaba-examples/pom.xml
```diff
@@ -56,6 +56,27 @@
 		<module>integrated-example/integrated-frontend</module>
 	</modules>
 
+    <properties>
+        <seata.version>1.6.1-SNAPSHOT</seata.version>
+        <nacos-client.version>1.4.2</nacos-client.version>
+    </properties>
+
+    <dependencyManagement>
```
steverao: It just temporary dependencies, we should don't add them in the PR. Maybe when we need to test and we add them manually.

steverao (approved):
LGTM, thanks @ruansheng8

PR #2989, opened 2022-12-20
Fix bootstrap bug for integrated example

Phixsura:
Fix bootstrap bug for integrated example


steverao (approved):
LGTM, thanks @MieAh Please create a same PR to branch of 2022.x

PR #3009, opened 2022-12-25
make PostConstruct method public

ruansheng8:
When using `AOT` build, the method marked by the annotation `@PostConstruct` must be public


steverao (approved):
LGTM, thanks @ruansheng8

PR #3045, opened 2023-01-05
Fix the bug on delete Label-routing rule

123liuziming:
1. Polish doc of governance module
2. Fix the bug on delete label-routing rule. I think the originalRouteData filed will prevent the deletion of label-routing rule, so we just remove the field. And we will update the routeCache and pathRuleMap atomically.


steverao:
@HHoflittlefish777 Please review it.

sollhui:
Part of Label-routing had been reviewed

PR #3070, opened 2023-01-07
update seata version

yuluo-yx:
update seate version to 1.6.1 in 2021.x branch

### DIFF
--- a/spring-cloud-alibaba-dependencies/pom.xml
+++ b/spring-cloud-alibaba-dependencies/pom.xml
@@ -20,7 +20,7 @@
     <properties>
         <revision>2021.0.4.0-SNAPSHOT</revision>
         <sentinel.version>1.8.5</sentinel.version>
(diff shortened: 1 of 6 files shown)


steverao:
Do you run the example module to verify the compatibility?

yuluo-yx:
> Do you run the example module to verify the compatibility?

Yes. I used NACOS with 2.2.0 locally, and the services were successfully registered and accessed the test controller, and I got the expected results. They seem to fit, at least locally in my area!

yuluo-yx:
I got such a warning when installing dependencies
`[WARNING] The POM for com.alibaba:druid:jar:1.2.6 is invalid, transitive dependencies (if any) will not be available, enable debug logging for more details`
This should be an issue with druid, but it doesn't affect use. So I ignored the warning.

yuluo-yx:
> Do you run the example module to verify the compatibility?

Please wait a minute, I found the problem and I'm fixing it

PR #3073, opened 2023-01-10
Polish code & fix envoy control plane api dependency version

123liuziming:
Polish code 
fix envoy control plane api dependency version
delete fastjson2 dependency in test


steverao (approved):
LGTM, thanks @123liuziming

Issue #2971, opened 2022-12-13
Use new features of jdk17 to optimize code in 2022.x branch.


Phixsura:
### BackGround
SpringCloudAlibaba's 2022.x branch is adapted to SpringBoot3 and jdk17. The leap between JDK8 and JDK17 has many new features, such as instanceof pattern matching, switch expressions, record classes, etc., which can reduce the amount of code, reduce maintenance costs, and improve performance. Next, I'll list new features that I think can optimize the 2022.x branch.

### instanceof pattern matching
The instanceof operator in Java is used to check the type of an object. The following code shows the code paradigm that typically uses the instanceof operator. Use instanceof in the if statement for checking, and if the check passes, use a cast to convert the input object obj to a String str, and finally use the variable str.
![image](https://user-images.githubusercontent.com/100568933/207393988-729719ce-34f9-427f-9a94-1c7791ea09fe.png)
As you can see from the above code, the paradigm of using the instanceof operator is very cumbersome, and the target type String that needs to be checked appears three times. With instanceof pattern matching, the code can be simplified a lot. In the following code, Strings represents the type pattern, where String is the type that needs to be matched and str is the variable used to capture the target object after the match is successful. The variable str can be used directly in an if statement block.
![image](https://user-images.githubusercontent.com/100568933/207394401-20b34fb1-5d70-4232-bf38-c5138a72a9d1.png)

### switch expressions
In JDK8 traditional switch has a long-term problem that has plagued users, Java's switch closely follows C and C++ languages, and supports fall through control by default. And the reason why the motivation for switching is enhanced is that while this traditional flow of control is often useful for writing low-level code (such as binary-encoded parsers), as switch is used in high-level contexts, its error-prone nature is starting to outstrip its flexibility. Many break statements make it unnecessarily verbose, and this visual noise often masks errors that are difficult to debug, where the absence of break statements means unexpected failure. As the following code example
![image](https://user-images.githubusercontent.com/100568933/207397371-4a7e5d49-194f-42a3-b2db-ce5c0e5b2b36.png)
Now the JDK introduces a new form of Switch tag "case L ->" to indicate that if the tag matches, only the code to the right of the tag is executed. Each case also allows multiple constants, separated by commas. The previous code can now be written as:
![image](https://user-images.githubusercontent.com/100568933/207398409-4b1c73ae-89a3-4be6-af6f-d687004f5af9.png)
[...]

PR #3095, opened 2023-01-13
Use methods inside Objects to override 'Obj == null' and 'Obj != null' and 'obj == obj'

yuluo-yx:
Canonical code format! #2971


yuluo-yx:
WIR

PR #3109, opened 2023-01-19
fix: fix the bug that the eds protocol is never observed

123liuziming:
The judging condition in PilotExchanger::observeClusters should be reversed.


ruansheng8 (approved):

steverao (approved):
LGTM.

Issue #3075, opened 2023-01-10
Add out-of-box support for Sentinel OpenSergo data-source


sczyh30:
**Which Component**

Sentinel, OpenSergo

**Describe the solution you'd like**

Sentinel v1.8.6 introduces [the data-source module](https://github.com/alibaba/Sentinel/tree/master/sentinel-extension/sentinel-datasource-opensergo) of [OpenSergo fault-tolerance spec](https://github.com/opensergo/opensergo-specification/blob/main/specification/en/fault-tolerance.md). We could add support for OpenSergo data-source in `spring-cloud-starter-alibaba-sentinel` module (perhaps in `spring-cloud-alibaba-sentinel-datasource`).

Sentinel 1.8.6 版本引入了 [OpenSergo 数据源模块](https://sentinelguard.io/zh-cn/docs/opensergo-data-source.html)，从而支持通过 [OpenSergo 流量防护与容错 spec CRD](https://github.com/opensergo/opensergo-specification/blob/main/specification/zh-Hans/fault-tolerance.md) 来统一管控流量防护规则。SCA 里面 Sentinel 相关 starter 模块可以添加对该模块的快速接入支持，以便用户可以可以很方便地通过 OpenSergo 统一控制面来管控 SCA 的流量治理规则。

**Additional context**

OpenSergo 流量防护与容错 spec: https://github.com/opensergo/opensergo-specification/blob/main/specification/zh-Hans/fault-tolerance.md

123liuziming:
Please assign it to me.

steverao:
It sounds good!

sczyh30:
一个简单的用户使用流程示例（仅做流程参考，名称、形式后续讨论）：

用户引入 SCA Sentinel starter 模块，按照以下配置即可快速开启 OpenSergo spec 支持，无需再改代码：

```
# 配置 OpenSergo 控制面 endpoint
spring.cloud.opensergo.endpoint=127.0.0.1:10246

spring.cloud.sentinel.datasource.opensergo.enabled-rules=flow,degrade,system
```

123liuziming:
I‘ve submit a POC pr here https://github.com/alibaba/spring-cloud-alibaba/pull/3142

PR #3142, opened 2023-02-06
feat: Add support for sentinel opensergo datasource

123liuziming:
Add out-of-box support for Sentinel OpenSergo data-source.

If you want to use OpenSergo datasource, you can configure the `application.yml` like NacosDatasource in Spring Cloud Alibaba:
```
# you can set single rule type by setting enabled-rules, like the other datasource in Spring Cloud Alibaba
spring.cloud.sentinel.datasource.ds5.opensergo.rule-type=flow
spring.cloud.sentinel.datasource.ds5.opensergo.host=127.0.0.1
spring.cloud.sentinel.datasource.ds5.opensergo.port=10246
spring.cloud.sentinel.datasource.ds5.opensergo.namespace=ns1
spring.cloud.sentinel.datasource.ds5.opensergo.app=app1
# you can set multiple rule type by setting enabled-rules
spring.cloud.sentinel.datasource.ds5.opensergo.enabled-rules[0]=flow
spring.cloud.sentinel.datasource.ds5.opensergo.enabled-rules[0]=degrade
```
#3075

### DIFF
--- a/spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/pom.xml
+++ b/spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/pom.xml
@@ -14,6 +14,10 @@
     <artifactId>spring-cloud-alibaba-sentinel-datasource</artifactId>
     <name>Spring Cloud Alibaba Sentinel DataSource</name>
 
+    <properties>
+        <opensergo-datasource.version>2.0.0-SNAPSHOT</opensergo-datasource.version>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>com.alibaba.cloud</groupId>
@@ -79,6 +83,12 @@
             <groupId>com.alibaba.csp</groupId>
             <artifactId>sentinel-datasource-consul</artifactId>
             <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>com.alibaba.csp</groupId>
+            <artifactId>sentinel-datasource-opensergo</artifactId>
+            <version>${opensergo-datasource.version}</version>
         </dependency>
 
         <dependency>
--- a/spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/pom.xml
+++ b/spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/pom.xml
@@ -15,7 +15,7 @@
     <name>Spring Cloud Alibaba Sentinel DataSource</name>
 
     <properties>
-        <opensergo-datasource.version>2.0.0-SNAPSHOT</opensergo-datasource.version>
+        <opensergo-datasource.version>0.1.0-beta</opensergo-datasource.version>
     </properties>
 
     <dependencies>
--- a/spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/OpenSergoDataSourceProperties.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/OpenSergoDataSourceProperties.java
@@ -38,7 +38,8 @@ import org.springframework.util.CollectionUtils;
  * @author <a href="liuziming@buaa.edu.cn"></a>
  */
 public class OpenSergoDataSourceProperties extends AbstractDataSourceProperties {
-
+	private static final String FLOW = "flow";
+	private static final String DEGRADE = "degrade";
 	private String host = "127.0.0.1";
 
 	private int port = 10246;
@@ -55,10 +56,10 @@ public class OpenSergoDataSourceProperties extends AbstractDataSourceProperties 
 
 	public void postRegister(OpenSergoDataSourceGroup dataSourceGroup) {
 		// TODO: SystemRule and ParamFlowRule
-		if (enabledRules.contains(OpenSergoSentinelConstants.KIND_PARAM_FLOW_RULE)) {
+		if (enabledRules.contains(FLOW)) {
 			FlowRuleManager.register2Property(dataSourceGroup.subscribeFlowRules());
 		}
-		if (enabledRules.contains(OpenSergoSentinelConstants.KIND_CIRCUIT_BREAKER_RULE)) {
+		if (enabledRules.contains(DEGRADE)) {
 			DegradeRuleManager.register2Property(dataSourceGroup.subscribeDegradeRules());
 		}
 		// When there is no enabled-rules, try ruleType
--- a/spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/OpenSergoDataSourceProperties.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/OpenSergoDataSourceProperties.java
@@ -16,6 +16,7 @@
 
 package com.alibaba.cloud.sentinel.datasource.config;
 
+import java.util.HashSet;
 import java.util.Set;
 
 import com.alibaba.cloud.commons.lang.StringUtils;
@@ -46,7 +47,7 @@ public class OpenSergoDataSourceProperties extends AbstractDataSourceProperties 
 
 	private String app = AppNameUtil.getAppName();
 
-	private Set<String> enabledRules;
+	private Set<String> enabledRules = new HashSet<>();
 
 	public OpenSergoDataSourceProperties() {
 		super(OpenSergoDataSourceFactoryBean.class.getName());
--- a/spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/OpenSergoDataSourceProperties.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/OpenSergoDataSourceProperties.java
@@ -25,6 +25,7 @@ import com.alibaba.csp.sentinel.datasource.OpenSergoDataSourceGroup;
 import com.alibaba.csp.sentinel.datasource.OpenSergoSentinelConstants;
 import com.alibaba.csp.sentinel.slots.block.degrade.DegradeRuleManager;
 import com.alibaba.csp.sentinel.slots.block.flow.FlowRuleManager;
+import com.alibaba.csp.sentinel.util.AppNameUtil;
 
 import org.springframework.util.CollectionUtils;
 
@@ -43,7 +44,7 @@ public class OpenSergoDataSourceProperties extends AbstractDataSourceProperties 
 
 	private String namespace = "default";
 
-	private String app;
+	private String app = AppNameUtil.getAppName();
 
 	private Set<String> enabledRules;
 
@@ -68,17 +69,6 @@ public class OpenSergoDataSourceProperties extends AbstractDataSourceProperties 
 		case DEGRADE:
 			DegradeRuleManager.register2Property(dataSourceGroup.subscribeDegradeRules());
 			break;
-		}
-	}
-
-	@Override
-	public void preCheck(String dataSourceName) {
-		if (StringUtils.isEmpty(app)) {
-			throw new IllegalArgumentException("OpenSergoDataSource app is empty");
-		}
-		if (CollectionUtils.isEmpty(enabledRules)) {
-			throw new IllegalArgumentException(
-					"OpenSergoDataSource enabled-rules is empty");
 		}
 	}
 
--- a/spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/OpenSergoDataSourceProperties.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/OpenSergoDataSourceProperties.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright 2013-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.alibaba.cloud.sentinel.datasource.config;
+
+import java.util.Set;
+
+import com.alibaba.cloud.commons.lang.StringUtils;
+import com.alibaba.cloud.sentinel.datasource.RuleType;
+import com.alibaba.cloud.sentinel.datasource.factorybean.OpenSergoDataSourceFactoryBean;
+import com.alibaba.csp.sentinel.datasource.OpenSergoDataSourceGroup;
+import com.alibaba.csp.sentinel.datasource.OpenSergoSentinelConstants;
+import com.alibaba.csp.sentinel.slots.block.degrade.DegradeRuleManager;
+import com.alibaba.csp.sentinel.slots.block.flow.FlowRuleManager;
+
+import org.springframework.util.CollectionUtils;
+
+/**
+ * OpenSergo Properties class Using by {@link DataSourcePropertiesConfiguration} and
+ * {@link OpenSergoDataSourceFactoryBean}.
+ *
+ * @author musi
+ * @author <a href="liuziming@buaa.edu.cn"></a>
+ */
+public class OpenSergoDataSourceProperties extends AbstractDataSourceProperties {
+
+	private String host = "127.0.0.1";
+
+	private int port = 10246;
+
+	private String namespace = "default";
+
+	private String app;
+
+	private Set<String> enabledRules;
+
+	public OpenSergoDataSourceProperties() {
+		super(OpenSergoDataSourceFactoryBean.class.getName());
+	}
+
+	public void postRegister(OpenSergoDataSourceGroup dataSourceGroup) {
+		// TODO: SystemRule and ParamFlowRule
+		if (enabledRules.contains(OpenSergoSentinelConstants.KIND_PARAM_FLOW_RULE)) {
+			FlowRuleManager.register2Property(dataSourceGroup.subscribeFlowRules());
+		}
+		if (enabledRules.contains(OpenSergoSentinelConstants.KIND_CIRCUIT_BREAKER_RULE)) {
+			DegradeRuleManager.register2Property(dataSourceGroup.subscribeDegradeRules());
+		}
+		// When there is no enabled-rules, try ruleType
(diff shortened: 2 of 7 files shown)


spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/pom.xml
```diff
@@ -14,6 +14,10 @@
     <artifactId>spring-cloud-alibaba-sentinel-datasource</artifactId>
     <name>Spring Cloud Alibaba Sentinel DataSource</name>
 
+    <properties>
+        <opensergo-datasource.version>2.0.0-SNAPSHOT</opensergo-datasource.version>
```
sczyh30: We could use `0.1.0-beta` version here.
Refer https://opensergo.io/zh-cn/docs/quick-start/java/quick-start-sentinel/

123liuziming: Yeah, I will fix it

spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/OpenSergoDataSourceProperties.java
```diff
@@ -0,0 +1,125 @@
...
+			break;
+		case DEGRADE:
+			DegradeRuleManager.register2Property(dataSourceGroup.subscribeDegradeRules());
+			break;
+		}
+	}
+
+	@Override
+	public void preCheck(String dataSourceName) {
+		if (StringUtils.isEmpty(app)) {
```
sczyh30: 这里面是否可以简化使用方式：如果 app 没有配，则按照 SCA Sentinel 取 appName 的方式来取？

另外这些 check 是否当且仅当 OpenSergo endpoint 有实际配置时才生效？

spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/pom.xml
```diff
@@ -81,6 +85,12 @@
             <optional>true</optional>
         </dependency>
 
+        <dependency>
+            <groupId>com.alibaba.csp</groupId>
+            <artifactId>sentinel-datasource-opensergo</artifactId>
+            <version>${opensergo-datasource.version}</version>
+        </dependency>
```
ruansheng8, line 92: Whether to use` optional`, the integration of `opensergo` is at the discretion of the user.

123liuziming: Without this dependency, I can not use the OpenSergoDataSourceGroup

spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/OpenSergoDataSourceProperties.java
123liuziming: 1. 可以
2. 应该不需要这个preCheck逻辑了，现在所有的属性应该都有默认值

sczyh30:
Could you please provide a sample `application.properties` config for the Sentinel OpenSergo data-source in the PR description?

123liuziming:
> Could you please provide a sample `application.properties` config for the Sentinel OpenSergo data-source in the PR description?

done

spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/OpenSergoDataSourceProperties.java
```diff
@@ -0,0 +1,117 @@
...
+	private static final String DEGRADE = "degrade";
+	private String host = "127.0.0.1";
+
+	private int port = 10246;
+
+	private String namespace = "default";
+
+	private String app = AppNameUtil.getAppName();
+
+	private Set<String> enabledRules = new HashSet<>();
```
jnan806, line 51: Maybe can use `Set<RuleType>`

123liuziming: Yeah, It is better to use Enum type. I will fix it!

spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/OpenSergoDataSourceProperties.java
```diff
@@ -0,0 +1,117 @@
...
+		// When there is no enabled-rules, try ruleType
+		RuleType ruleType = getRuleType();
+		switch (ruleType) {
+		case FLOW:
+			FlowRuleManager.register2Property(dataSourceGroup.subscribeFlowRules());
+			break;
+		case DEGRADE:
+			DegradeRuleManager.register2Property(dataSourceGroup.subscribeDegradeRules());
+			break;
+		}
```
jnan806, line 74: As your comment, these code lines should be wrapped in condition `enabledRules`  is empty. 
Because the `ruleType` was annotationed by `NotNull` in `AbstractDataSourceProperties` ，so if `enabledRules ` contains the `ruleType`, the `subscribeConfig` action will be invoked twice.

oss-sentinel-ai (commented):
### Summary
This PR adds a new Sentinel datasource backed by [OpenSergo](https://opensergo.io/), so flow and degrade rules can be sourced from an OpenSergo control plane instead of the existing file / Apollo / Zookeeper / Redis / Consul / Nacos backends. It introduces `OpenSergoDataSourceProperties` and `OpenSergoDataSourceFactoryBean` in `spring-cloud-alibaba-sentinel-datasource`, wires a new `opensergo` field into `DataSourcePropertiesConfiguration`, extends `SentinelDataSourceHandler.registerBean` with a branch that handles the fact that `OpenSergoDataSourceGroup` does *not* extend `AbstractDataSource`, and registers `opensergo = com.alibaba.csp.sentinel.datasource.OpenSergoDataSourceGroup` in the `sentinel-datasource.properties` SPI map. An example property and a non-optional Maven dependency on `sentinel-datasource-opensergo:0.1.0-beta` round out the change.

The idea is reasonable, but the implementation has a real correctness bug and a handful of conventions/quality issues that should be addressed before merging. Most importantly, `OpenSergoDataSourceProperties.postRegister` can **double-register** the same rule property with Sentinel's `FlowRuleManager` / `DegradeRuleManager` when both `enabledRules` and the legacy `ruleType` resolve to the same kind (the `ruleType` `switch` is *not* an `else` of the `enabledRules` branch, despite the code comment claiming it is). In addition, the handler branch silently swallows startup failures, the new Maven dependency is non-optional (breaking the convention used by every other datasource binding in this module), the factory bean exposes unused `converter` / `enabledRules` setters, and there are no tests. The example also stacks a `file` datasource and an `opensergo` datasource under the same `ds5` name, which is misleading.

## Findings

- **Critical** `OpenSergoDataSourceProperties.java:30` — `postRegister` double-registers flow/degrade rules when both `enabledRules` and `ruleType` match the same kind; the `ruleType` `switch` should be an `else` branch, and the `switch` silently drops `PARAM_FLOW` / `SYSTEM` / `AUTHORITY` despite the example advertising `param_flow`.
- **Warning** `SentinelDataSourceHandler.java:351` — `catch (Exception e)` swallows `dataSourceGroup.start()` and `postRegister` failures with only a generic log line; startup failures should surface loudly, or at minimum log the datasource name and target endpoint.
- **Warning** `pom.xml:29` — `sentinel-datasource-opensergo` is declared non-optional and without scope; every other datasource binding in this module is `<optional>true</optional>`.
- **Warning** `OpenSergoDataSourceFactoryBean.java:43` — `getObject()` passes `host` / `port` / `namespace` / `app` into the group constructor without null/blank/range validation; bad user input fails silently at subscribe time instead of at startup.
- **Warning** `OpenSergoDataSourceProperties.java:20` — `private String app = AppNameUtil.getAppName();` resolves at properties-construction time, before the Spring environment is ready, so `app` can get stuck as `null`.
- **Warning** `OpenSergoDataSourceFactoryBean.java:40` — `converter` and `enabledRules` fields are settable but never read by `getObject()`; dead API surface.
- **Warning** `OpenSergoDataSourceProperties.java:29` — `// TODO: SystemRule and ParamFlowRule` ships alongside an example that sets `rule-type=param_flow`, a combination that silently does nothing.
- **Warning** `application.properties:40` — the example has both `ds5.file.*` and `ds5.opensergo.*` under the same datasource name; `file` and `opensergo` are sibling fields in `DataSourcePropertiesConfiguration`, so this is confusing. The existing `file.file=classpath: param-flow.json` also contains a suspicious space after `classpath:`.
- **Info** `pom.xml:19` — `opensergo-datasource.version` is pinned to `0.1.0-beta`; pre-release dependencies are unusual for a GA spring-cloud-alibaba release.
- **Info** `SentinelDataSourceHandler.java:335` — the cast to `Object` is required, but there is no fallback `else` for datasource types that are neither `AbstractDataSource` nor `OpenSergoDataSourceGroup`; silent no-op makes future additions risky.
- **Info** `OpenSergoDataSourceProperties.java:11` — no tests cover the new properties, factory bean, handler branch, or the double-registration path; please add coverage consistent with the other datasource modules.
- **Info** `OpenSergoDataSourceProperties.java:111` — `@author <a href="liuziming@buaa.edu.cn"></a>` is malformed HTML that renders as an empty anchor; use `mailto:` or plain text.
- **Info** `OpenSergoDataSourceProperties.java:11` — new configuration keys (`host`, `port`, `namespace`, `app`, `enabled-rules`, `rule-type`) are introduced without a companion docs PR.

## Branch & docs notes

- The target branch, **`2.2.x-ospp2023`**, was a one-off branch for the 2023 OSPP mentorship program and is **no longer maintained**; the currently maintained branches are `2025.1.x` / `2025.0.x` / `2023.x`. Before investing further in this PR, the author should confirm with the maintainers whether the OpenSergo integration is still wanted and, if so, **rebase it onto a maintained branch** (likely `2023.x`, given the feature-add nature of the change) — otherwise this will not be shipped.
- Because new public configuration properties are added, please also open a companion PR against [spring-cloud-alibaba-group/spring-cloud-alibaba-group.github.io](https://github.com/alibaba/spring-cloud-alibaba-group/spring-cloud-alibaba-group.github.io) documenting the `opensergo` datasource keys, their defaults, and the supported `rule-type` values.

### Findings
- **[CRITICAL]** `spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/OpenSergoDataSourceProperties.java:30` — Bug: `postRegister` can double-register rules with Sentinel's `FlowRuleManager` / `DegradeRuleManager`. If `enabledRules` contains `flow` (or `degrade`) AND `getRuleType()` is also `FLOW` (or `DEGRADE`), the same `dataSourceGroup.subscribeFlowRules()` / `subscribeDegradeRules()` property is registered twice, producing two concurrent subscribers on the same OpenSergo topic and two writes into the same `SentinelProperty`. Either make the `ruleType` switch an `else` branch that only runs when `enabledRules` is empty (matching the 'When there is no enabled-rules, try ruleType' comment), or dedupe with a set of already-registered rule kinds. Also: the TODO on line 29 and the incomplete `switch` (no `PARAM_FLOW` / `SYSTEM` / `AUTHORITY` / `PARAMETER_FLOW`, despite the example advertising `param_flow`) mean the fallback branch silently does nothing for those rule types — at minimum log a warning.
- **[WARNING]** `spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-sentinel/src/main/java/com/alibaba/cloud/sentinel/custom/SentinelDataSourceHandler.java:351` — The `catch (Exception e)` swallows failures from `dataSourceGroup.start()` and `postRegister(...)`, only logging them. Other datasource paths let registration errors propagate (which surfaces misconfig fast); swallowing here means a user who typos `host`/`port`/`namespace` gets a silent no-op and a single log line that is easy to miss. Consider either re-throwing as a `BeanCreationException` (or `IllegalStateException`) so startup fails loudly, or at least logging the datasource name (`dataSourceName`) and the target `host:port/namespace` so the offending bean is identifiable in a multi-datasource deployment.
- **[WARNING]** `spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/pom.xml:29` — `sentinel-datasource-opensergo` is declared as a non-optional, unscoped (compile) dependency. Every other datasource binding in this module (Apollo, Zookeeper, Redis, Consul, …) is `<optional>true</optional>` so that users only pull in the one they actually configure. As written, every consumer of `spring-cloud-alibaba-sentinel-datasource` transitively pulls the OpenSergo client even if they never use it. Please add `<optional>true</optional>` to match the existing convention.
- **[WARNING]** `spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/factorybean/OpenSergoDataSourceFactoryBean.java:43` — `getObject()` passes `host`, `namespace`, and `app` straight into the `OpenSergoDataSourceGroup` constructor with no validation. The property defaults are fine, but Spring binding can overwrite them with `null` or an empty string (e.g. `spring.cloud.sentinel.datasource.dsX.opensergo.host=`), which would silently produce a broken client that only fails at subscribe time. Validate non-null/non-blank `host`, `namespace`, and `app` and a valid `port` range here (or in `OpenSergoDataSourceProperties`) and throw `IllegalArgumentException` with a message that names the offending property.
- **[WARNING]** `spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/OpenSergoDataSourceProperties.java:20` — `private String app = AppNameUtil.getAppName();` evaluates at properties-construction time. `AppNameUtil.getAppName()` reads `project.name` / `spring.application.name` from a static `SentinelConfig` which is populated later in the Spring lifecycle; depending on bean creation order, this can resolve to `null` at binding time and then get stuck as `null` even after the environment is ready. Either resolve `app` lazily in the factory bean (read `Environment.getProperty("spring.application.name")` if `app` is null/blank), or at least document that users must set `app` explicitly.
- **[WARNING]** `spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/factorybean/OpenSergoDataSourceFactoryBean.java:40` — The `converter` field has a getter/setter but is never used: `getObject()` constructs `new OpenSergoDataSourceGroup(host, port, namespace, app)` and ignores `converter`, and `enabledRules` is also copied onto the bean but unused by `getObject()` (it is only consumed by `OpenSergoDataSourceProperties.postRegister`). Dead setters on a `FactoryBean` are misleading — other datasource factory beans (e.g. `ApolloDataSourceFactoryBean`) actually wire the converter through. Either drop the unused fields, or actually use them (and document which rule types the converter applies to).
- **[WARNING]** `spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/OpenSergoDataSourceProperties.java:29` — The `// TODO: SystemRule and ParamFlowRule` comment is shipping as part of the feature, yet the example in `application.properties` advertises `spring.cloud.sentinel.datasource.ds5.opensergo.rule-type=param_flow` — a combination that will silently do nothing because neither the `enabledRules` branch nor the `switch` handles `PARAM_FLOW`. Either finish the support in this PR, or drop the TODO and the misleading example, and log a warning when an unsupported `ruleType` is requested.
- **[WARNING]** `spring-cloud-alibaba-examples/sentinel-example/sentinel-core-example/src/main/resources/application.properties:40` — This example now has both `spring.cloud.sentinel.datasource.ds5.file.*` AND `spring.cloud.sentinel.datasource.ds5.opensergo.*` under the same datasource name `ds5`. In `DataSourcePropertiesConfiguration`, `file` and `opensergo` are sibling fields — they don't nest — so `ds5` would bind whichever property set the handler picks, not both, which is confusing for readers of the example. Additionally, the existing `file.file=classpath: param-flow.json` contains a suspicious space after `classpath:` (likely a typo for `classpath:param-flow.json`). Please either give the OpenSergo example its own datasource name (e.g. `ds6`) with a commented explanation, or split it into a separate example properties file.
- **[INFO]** `spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/pom.xml:19` — `opensergo-datasource.version` is pinned to `0.1.0-beta`. Beta artifacts are not normally referenced from a GA spring-cloud-alibaba release — they carry no stability guarantee and may change coordinates/semantics between betas. If OpenSergo is now GA, bump to a stable release; otherwise, call out the pre-release status in the PR description and consider marking the whole integration as experimental (e.g. an `@Experimental` annotation or a package-info note) so downstream users know.
- **[INFO]** `spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-sentinel/src/main/java/com/alibaba/cloud/sentinel/custom/SentinelDataSourceHandler.java:335` — The cast from `AbstractDataSource` to `Object` is needed because `OpenSergoDataSourceGroup` doesn't extend `AbstractDataSource`, but the follow-up `instanceof OpenSergoDataSourceGroup` branch does not exclude that other datasource types could also be neither `AbstractDataSource` nor `OpenSergoDataSourceGroup` — in which case the method silently does nothing (no log, no exception). Consider adding a final `else` that logs a warning like `"Unsupported datasource type for '{}': {}"` to make future additions safer and misconfigurations visible.
- **[INFO]** `spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/OpenSergoDataSourceProperties.java:11` — No unit or integration tests are included for `OpenSergoDataSourceProperties`, `OpenSergoDataSourceFactoryBean`, the new `postRegister` branch, or the double-registration bug path. The existing Sentinel datasource module has tests under `src/test/java` covering each factory bean and the handler's registration flow — please add analogous tests here (at minimum: property binding round-trip; factory bean produces a group with the configured host/port/namespace/app; handler happy path with a mocked `OpenSergoDataSourceGroup`; and a test proving the new `instanceof OpenSergoDataSourceGroup` branch triggers `start()` + `postRegister`). Given the subtle double-registration bug, a regression test for that specific case would be valuable.
- **[INFO]** `spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/OpenSergoDataSourceProperties.java:111` — Author tag is malformed: `@author <a href="liuziming@buaa.edu.cn"></a>` is raw HTML inside a Javadoc `@author` line and renders as an empty anchor. Use a plain form like `@author <a href="mailto:liuziming@buaa.edu.cn">Ziming Liu</a>` or just `@author Ziming Liu`. Also `@author musi` appears in both this file and the factory bean — confirm that's the intended display name and that it matches the project's contributor conventions.
- **[INFO]** `spring-cloud-alibaba-starters/spring-cloud-alibaba-sentinel-datasource/src/main/java/com/alibaba/cloud/sentinel/datasource/config/OpenSergoDataSourceProperties.java:11` — Several new configuration keys are introduced (`spring.cloud.sentinel.datasource.<name>.opensergo.host|port|namespace|app|enabled-rules|rule-type`) but no docs changes accompany this PR. Spring Cloud Alibaba documents configuration keys in the sibling `spring-cloud-alibaba-group/spring-cloud-alibaba-group.github.io` repository — please open (or link) a companion docs PR describing the new `opensergo` datasource, the supported `rule-type` values, the meaning of `enabled-rules`, and the default `port=10246`/`namespace=default`/`host=127.0.0.1` so operators know what to set.

---

PR #3160, opened 2023-02-18
Fix nacos discovery/config log 

### LINKED ISSUE #3133: Nacos-Client log print to application logs, not print to nacos' log
**Which Component**
Nacos Discovery/Config

**Describe the bug**
Relative issue #1931.

After the issue fix problem, the problem fixed by some old version.

But in some new version, the problem started again. After my debug and research. I fount that the problem is caused by the spring will call the `LoggingSystem.initialize` from `PropertySourceBootstrapConfiguration` after all `ApplicationEnvironmentPreparedEvent`.

But `NacosLoggingListener` in Nacos Discovery/Config only handle `ApplicationEnvironmentPreparedEvent` not loadConfiguration again after `PropertySourceBootstrapConfiguration`. So the log config still be overwrite by spring.

**To Reproduce**
1. Go to 'nacos-discovery-spring-cloud-config-client-example' in tag `2.2.9.RELEASE`
2. Run on 'SpringCloudConfigClientApplication'
3. See nacos client log print into console.

**Expected behavior**
nacos client log print into ${user.home}/naming.log or config.log 

**Screenshots**

**Additional context**
MacOS  、Java8 、 Version 2.2.6.RELEASE ~ 2.2.9.RELEASE

I don't test for 2021.X and 2022.X. Please community to check and fix.

ruansheng8:
Fixed #3133

### DIFF
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/logging/NacosLoggingAppRunListener.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/logging/NacosLoggingAppRunListener.java
@@ -18,38 +18,36 @@ package com.alibaba.cloud.nacos.logging;
 
 import com.alibaba.nacos.client.logging.NacosLogging;
 
-import org.springframework.boot.context.event.ApplicationEnvironmentPreparedEvent;
-import org.springframework.context.ApplicationEvent;
-import org.springframework.context.event.GenericApplicationListener;
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.SpringApplicationRunListener;
+import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.core.Ordered;
-import org.springframework.core.ResolvableType;
+import org.springframework.core.env.ConfigurableEnvironment;
 
 /**
- * Reload nacos log configuration file, after
- * {@link org.springframework.boot.context.logging.LoggingApplicationListener}.
+ * Reload nacos log configuration on spring application contextPrepared.
  *
- * @author mai.jh
+ * @author RuanSheng
  */
-@Deprecated
-public class NacosLoggingListener implements GenericApplicationListener {
+public class NacosLoggingAppRunListener implements SpringApplicationRunListener, Ordered {
 
-	@Override
-	public boolean supportsEventType(ResolvableType resolvableType) {
-		Class<?> type = resolvableType.getRawClass();
-		if (type != null) {
-			return ApplicationEnvironmentPreparedEvent.class.isAssignableFrom(type);
-		}
-		return false;
+	public NacosLoggingAppRunListener(SpringApplication application, String[] args) {
+
 	}
 
 	@Override
-	public void onApplicationEvent(ApplicationEvent applicationEvent) {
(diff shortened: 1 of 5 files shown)


spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/logging/NacosLoggingListener.java
```diff
@@ -30,6 +30,7 @@
  *
  * @author mai.jh
  */
+@Deprecated
```
steverao, line 33: Whether we delete the invalid listener directly?

spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/main/java/com/alibaba/cloud/nacos/discovery/logging/NacosLoggingListener.java
```diff
@@ -30,6 +30,7 @@
  *
  * @author mai.jh
  */
+@Deprecated
```
steverao, line 33: Same problem like above.

spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/logging/NacosLoggingAppRunListener.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public class NacosLoggingAppRunListener implements SpringApplicationRunListener, Ordered {
+
+	public NacosLoggingAppRunListener(SpringApplication application, String[] args) {
+
+	}
+
+	@Override
+	public void environmentPrepared(ConfigurableEnvironment environment) {
+		NacosLogging.getInstance().loadConfiguration();
```
steverao, line 40: Whether we don't need do load configuration here? because I see it doesn't have clearing logContext logic before here and just have it before contextPrepared like following:
image

ruansheng8: The Nacos client will create connection after `environmentPrepared`, and log messages will be output to the console. If the `NacosLogging.getInstance().loadConfiguration()` is reset after `contextPrepared`, some of the log messages will be output to the console between `environmentPrepared` and `contextPrepared`.

spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/logging/NacosLoggingListener.java
ruansheng8: The reason why these two classes were not directly deleted is that there may be users who depend on these classes for some customization operations (although this possibility is small), so they are marked as `deprecated` first and will be removed in the next version.

steverao (approved):
LGTM, thanks @ruansheng8

ruansheng8:
`TODO:` I will remove `NacosLoggingListener` in the next version.

PR #3178, opened 2023-02-26
fix: remove invalid dependency define.

steve-altman:
Remove invalid dependency define of nacos-config.


steverao (approved):
LGTM, thanks @steve-altman

Issue #3202, opened 2023-03-10
Add docker support in SCA example


yuluo-yx:
**Which Component**
eg. `Spring Clouud Alibaba Example`

**Is your feature request related to a problem? Please describe.**
Add docker support to the sample module.
To eliminate tedious steps when you need to run multiple components locally at the same time.
For example, in the Seta-example, you need to start three components locally at the same time: mysql5.7, nacos2.1, and Seta-Server 1.5.0. The creation process is tedious.

**Describe the solution you'd like**
Add docker-compose support in the various example modules

steverao:
We completed it and close now.

PR #3209, opened 2023-03-15
add docker support for integreated-example

yuluo-yx:
add docker support for integrated-example.  #3202


steverao (approved):
LGTM, thanks @yuluo-yx

PR #3216, opened 2023-03-20
fix doc wrong

yuluo-yx:
fix doc wrong


steverao (approved):
LGTM, thanks @yuluo-yx

PR #3239, opened 2023-03-31
Fix: pollable consumer doesn't ack the message correctly

### LINKED ISSUE #3235: RocketMQ pollable-consumer never acknowledge(commit) message
**Which Component**
RocketMQ: 4.9.4
Spring Cloud Alibaba: 2021.x

**Describe the bug**
RocketMQ pollable-consumer never ack(commit) message, even if the message is consumed normally.

**Simplest demo**
The official example can reproduce the problem. 
[spring-cloud-alibaba-rocketmq-example](https://github.com/alibaba/spring-cloud-alibaba/tree/2022.x/spring-cloud-alibaba-examples/rocketmq-example/rocketmq-pollable-consume-example)

**To Reproduce**
1. run the above example. It will do two things: first produce 100 messages. Then consume them using pollable consumer.
2. Then re-run the example without producer, We can found that the messages will be consumed again, which means that the messages consumed NEVER COMMITED.

**Expected behavior**
The messages that are consumed normally used be acknowledged(committed)

**Screenshots**

LeBW:
Fix #3235 

### Describe how you did it
Ack the message (fix the wrong method invocation)

### Describe how to verify it
Using official example

### DIFF
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-stream-rocketmq/src/main/java/com/alibaba/cloud/stream/binder/rocketmq/integration/inbound/pull/RocketMQAckCallback.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-stream-rocketmq/src/main/java/com/alibaba/cloud/stream/binder/rocketmq/integration/inbound/pull/RocketMQAckCallback.java
@@ -26,6 +26,8 @@ import org.slf4j.LoggerFactory;
 import org.springframework.integration.acks.AcknowledgmentCallback;
 import org.springframework.util.Assert;
 
+import java.util.Collections;
+
 /**
  * A pollable {@link org.springframework.integration.core.MessageSource} for RocketMQ.
  *
@@ -79,7 +81,7 @@ public class RocketMQAckCallback implements AcknowledgmentCallback {
 				switch (status) {
 				case REJECT:
 				case ACCEPT:
-					consumer.committed(messageQueue);
+					consumer.commit(Collections.singleton(messageQueue), false);
 					break;
 				case REQUEUE:
 					consumer.seek(messageQueue, offset);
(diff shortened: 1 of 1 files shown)


spring-cloud-alibaba-starters/spring-cloud-starter-stream-rocketmq/src/main/java/com/alibaba/cloud/stream/binder/rocketmq/integration/inbound/pull/RocketMQAckCallback.java
```diff
@@ -79,7 +81,7 @@ public void acknowledge(Status status) {
 				switch (status) {
 				case REJECT:
 				case ACCEPT:
-					consumer.committed(messageQueue);
+					consumer.commit(Collections.singleton(messageQueue), false);
```
steverao, line 84: We need to evaluate whether it is appropriate to set the `persist` parameter here to `false`.

LeBW: > We need to evaluate whether it is appropriate to set the `persist` parameter here to `false`.

I confirmed the code and `persist` here should be 'false'. There are two reasons:
1. I check the code of `commitSync()`, which is a method in `Consumer` that commit all of the messages in all messageQueues of the topic. The behavior of the method is the same as `commit(messageQueue, false)` here.
2. Inside if the `Consumer` object, there is a scheduled task which will persist all consumer offset every 5s. So it's unnecessary to set the `persist` here to `true`.

steverao (approved):
LGTM, thanks @LeBW. BTW, Please check whether need to create same PR to other main branchs.

PR #3247, opened 2023-04-04
Fix: pollable consumer doesn't ack the message correctly

### LINKED ISSUE #3235: RocketMQ pollable-consumer never acknowledge(commit) message

LeBW:
fix #3235  in 2.2.x branch


steverao (approved):
LGTM, thanks @LeBW

PR #3280, opened 2023-04-22
Update the execution time of the action

yuluo-yx:
Update the execution time of the action


steverao (approved):
LGTM, thanks @yuluo-yx

PR #3324, opened 2023-05-26
When use the custom balancer and use the defaultServiceInstace throw …

zt9788:
When use the custom balancer and use the defaultServiceInstace throw ClassCaseException

None

### Describe how you did it
Add some judgments

### Describe how to verify it
None

### Special notes for reviews
None


zt9788:
name error to recommit agent

Issue #3375, opened 2023-07-09
[Bug] 2.2.x nacos-reactivediscovery-consumer-example run failed!


yuluo-yx:
**Which Component**
Nacos Discovery Example

**Describe the bug**
```
java.net.UnknownHostException: service-provider
	at java.net.Inet6AddressImpl.lookupAllHostAddr(Native Method) ~[na:1.8.0_131]
	Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException: 
Error has been observed at the following site(s):
	|_ checkpoint ⇢ Request to GET http://service-provider/echo/yuluo [DefaultWebClient]
	|_ checkpoint ⇢ Handler com.alibaba.cloud.examples.MyController#serviceCall(String) [DispatcherHandler]
	|_ checkpoint ⇢ org.springframework.boot.actuate.metrics.web.reactive.server.MetricsWebFilter [DefaultWebFilterChain]
	|_ checkpoint ⇢ HTTP GET "/service-call/yuluo" [ExceptionHandlingWebHandler]
Stack trace:
		at java.net.Inet6AddressImpl.lookupAllHostAddr(Native Method) ~[na:1.8.0_131]
		at java.net.InetAddress$2.lookupAllHostAddr(InetAddress.java:928) ~[na:1.8.0_131]
		at java.net.InetAddress.getAddressesFromNameService(InetAddress.java:1323) ~[na:1.8.0_131]
```

**Simplest demo**
https://github.com/alibaba/spring-cloud-alibaba/tree/2.2.x/spring-cloud-alibaba-examples/nacos-example/nacos-discovery-example/nacos-reactivediscovery-consumer-example

PR #3376, opened 2023-07-09
fix nacos-reactive-discovery-example run failed

yuluo-yx:

- Fix https://github.com/alibaba/spring-cloud-alibaba/issues/3375


steverao (approved):
LGTM, thanks @yuluo-yx

PR #3560, opened 2023-12-27
feat: add license header for yaml and yml file

yuluo-yx:

- Add license header for yaml and yml file

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,3 +1,19 @@
+#
+# Copyright 2013-2023 the original author or authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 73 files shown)


.circleci/config.yml
```diff
@@ -1,3 +1,19 @@
+#
+# Copyright 2013-2023 the original author or authors.
```
chickenlj, line 2: It's the year 2024 already 😀

chickenlj: But it's ok to update this part later together with other files of this project.

chickenlj (approved):

.circleci/config.yml
yuluo-yx: get, I will submit pr to update it when this pr merged.

PR #3625, opened 2024-02-15
feat: optimize sidecar-example and readme docs

yuluo-yx:

- Update sidecar-example docs, fix docs style error;
- Add golang-service example, node services are less used, golang is more informative.


ruansheng8 (approved):
LGTM.

PR #3669, opened 2024-04-01
example: add ai-example simple front pages and update ai-example code

yuluo-yx:
## Describe what this PR does / why we need it

1. Update sca-ai-example
    1. update ai-example backend code;
    2. add simple frontend pages.
2. add marked.js
    1. The marked.js version loaded by cdn was too high and had compatibility issues, so I chose to introduce it locally and directly into the.


chickenlj (approved):
LGTM.

PR #3827, opened 2024-08-14
fix: TongYiChatOptions params TopP

zhangqian9158:

fix: TongYiChatOptions params TopP


yuluo-yx (approved):
lgtm, thanks @zhangqian9158

PR #3871, opened 2024-10-22
2023.x support nacos config annotation split

shiyiyue1102:
support nacos config annotation.
@NacosConfig : apply to filed ,auto apply value from nacos to specific type filed, support primitive type,properties,object ,collections.
@NacosConfigListener: apply to method , auto notify value from nacos to method's paramized type, support primitive type,properties,object ,collections.
@NacosConfigKeyListener: apply to method, auto notify when key changed from properties or yaml config. accpect ConfigChangeItem param.
@NacosConfigKeyListener: apply to method, auto notify when matched keys changed from properties or yaml config. accpect ConfigChangeEvent param.

split spring-cloud-starter-alibaba-nacos-config to spring-cloud-starter-alibaba-nacos-config（with cloud api）and spring-alibaba-nacos-config（with boot api only），thus module  spring-alibaba-nacos-config can be use on SpingBoot Application


chickenlj (approved):
LGTM.

PR #3888, opened 2024-11-14
enhance nacos config annotation ,support applying to Spring Bean.

shiyiyue1102:

enhance nacos config annotation ,support applying to Spring Bean.


chickenlj (approved):
LGTM.

PR #3908, opened 2024-11-28
fix bug spring.cloud.nacos.config.enabled not work

shiyiyue1102:

fix bug spring.cloud.nacos.config.enabled not work


chickenlj (approved):
LGTM.

PR #4039, opened 2025-08-06
Enable @ConditionalOnLoadBalancerNacos by default, as Ribbon has been…

cuitlideng20-blip:

Avoid requiring users to manually configure `spring.cloud.loadbalancer.nacos.enabled=true` in app configuration files such as application.yml.
Instead, use Spring Boot auto-configuration to set up the load balancer.
Since the Spring Cloud Alibaba Nacos Discovery starter has removed Ribbon support, the only available load balancer is Spring Cloud LoadBalancer — it should be enabled by default.


uuuyuqi:
Hello, thank you for your PR! However, we currently do not recommend using Nacos LoadBalancer as the default option. This would change the default configuration for many users and could lead to some unexpected issues when integrating with SpringCloud LoadBalancer.

PR #4149, opened 2025-12-06
Fixbug issue4147 2025.1.x，同 PR 4148

xzxiaoshan:
解决问题 #4147，同PR  #4148

https://github.com/alibaba/spring-cloud-alibaba/pull/4148

合入 2025.1.x 分支。

### DIFF
--- a/spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/NacosConfigManager.java
+++ b/spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/NacosConfigManager.java
@@ -160,6 +160,11 @@ public class NacosConfigManager implements InitializingBean, EnvironmentAware {
 	 */
 	private void recreateConfigService(NacosConfigProperties properties) {
 		if (service != null) {
+			try {
+				service.shutDown();
+			} catch (NacosException e) {
+				log.warn("Failed to shutdown old Nacos ConfigService", e);
+			}
 			service = null;
 		}
 		createConfigService(properties);
--- a/spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/NacosConfigManager.java
+++ b/spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/NacosConfigManager.java
@@ -164,7 +164,8 @@ public class NacosConfigManager implements InitializingBean, EnvironmentAware {
 					INSTANCE.nacosConfigProperties = newProperties;
 					// Recreate ConfigService if needed
 					recreateConfigService(newProperties);
-				} else {
+				}
+				else {
 					log.debug("Skipping update of INSTANCE properties as it does not match current instance");
 				}
 			}
--- a/spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/NacosConfigManager.java
+++ b/spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/NacosConfigManager.java
@@ -56,7 +56,7 @@ public class NacosConfigManager implements InitializingBean, EnvironmentAware {
 	/*
 	 * Handler for binding configuration properties.
 	 */
-	private static BindHandler bindHandler;
+	private static volatile BindHandler bindHandler;
 
 	public NacosConfigManager(NacosConfigProperties nacosConfigProperties) {
 		this.nacosConfigProperties = nacosConfigProperties;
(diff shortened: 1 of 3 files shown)


spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/configdata/NacosConfigDataLocationResolver.java
```diff
@@ -117,7 +117,10 @@ protected Log getLog() {
 
 	@Override
 	public boolean isResolvable(ConfigDataLocationResolverContext context,
-			ConfigDataLocation location) {
+								ConfigDataLocation location) {
+		if (NacosConfigManager.getBindHandler() == null) {
+			NacosConfigManager.setBindHandler(this.getBindHandler(context));
```
kaori-seasons, line 122: `bindHandler` is statically writable. `NacosConfigDataLocationResolver` performs a `set` operation in `isResolvable`, which introduces a race condition (multiple concurrent parsing threads may repeatedly set or read a null value).

Suggestions:
Use atomic/double-checked locking for writing, or use `synchronized`/`compare-and-set` in `setBindHandler` to ensure it's initialized only once, or document that "the first setter is responsible for setting the value."

spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/NacosConfigManager.java
```diff
@@ -89,4 +109,84 @@ public NacosConfigProperties getNacosConfigProperties() {
...
+		if (Objects.nonNull(newProperties)) {
+			// Update current instance's properties
+			this.nacosConfigProperties = newProperties;
+			// Safely update singleton instance
+			if (INSTANCE != null) {
+				synchronized (NacosConfigManager.class) {
+					if (INSTANCE != null) {
+						INSTANCE.nacosConfigProperties = newProperties;
+						// Recreate ConfigService if needed
+						recreateConfigService(newProperties);
```
kaori-seasons: `recreateConfigService` only logs a warning when shutdown fails, then sets `service = null` and continues to call `createConfigService`. If `createConfigService` throws an exception, should we revert to the old instance? Currently, the old instance has already been attempted to be shut down and discarded. Consider logging more context or providing more detailed logs in case of failure.

spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/test/java/com.alibaba.cloud.nacos/endpoint/NacosConfigEndpointTests.java
```diff
@@ -81,6 +82,11 @@ private void checkoutAcmHealthIndicator() {
 		try {
 			Builder builder = new Builder();
 
+			ConfigService configService = properties.configServiceInstance();
+			// 因为 NacosConfigManager 的 afterPropertiesSet 中会重新创建 ConfigService，
+			// 导致当前测试用例类最开始 static 块中 Mockito 固定设置的 UP 失效，所以这里重新设置一下。
+			// 测试用例 static 块中固定设置的 UP，实际上为了跑测试用例，是假UP，这里重新设置没有逻辑影响。
+			Mockito.when(configService.getServerStatus()).thenReturn("UP");
```
kaori-seasons, line 89: tips: In testing, using `Mockito.when` to reset `getServerStatus` is a workaround for side effects, but a more robust approach is to avoid relying on the lifecycle of static singletons in tests.  This can be achieved by injecting a controllable `ConfigService` in the tests or refactoring the manager to support better test replacement points.

spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/NacosConfigManager.java
```diff
@@ -89,4 +109,84 @@ public NacosConfigProperties getNacosConfigProperties() {
...
+						.bind(nacosConfigPrefix, Bindable.ofInstance(properties), bindHandler)
+						.orElse(properties))
+				.orElseGet(() -> binder
+						.bind(nacosConfigPrefix, Bindable.of(NacosConfigDataLoadProperties.class), bindHandler)
+						.orElse(null));
+		if (Objects.nonNull(newProperties)) {
+			// Update current instance's properties
+			this.nacosConfigProperties = newProperties;
+			// Safely update singleton instance
+			if (INSTANCE != null) {
```
kaori-seasons: Problem: Global static variables can interfere with each other in testing or multi-Spring context scenarios.

Recommendation (medium-term):  Ideally, manage the lifecycle of services and instances using Spring (defining them as singleton beans) to avoid static state; if static variables are necessary, document their thread safety and reconstruction contract, and add synchronization strategies in the code.

Transitional solution: At a minimum, centralize the access and modification of static services within thread-safe methods and log these operations to avoid implicit concurrent modifications.

spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/configdata/NacosConfigDataLocationResolver.java
xzxiaoshan: Code has been adjusted

spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/test/java/com.alibaba.cloud.nacos/endpoint/NacosConfigEndpointTests.java
xzxiaoshan: 这里的测试用例还没有更好的方法，暂未改动。其他几点建议已经优化代码。

kaori-seasons (approved):
LGTM

PR #4245, opened 2026-02-04
fix: workaround for Seata autoconfiguration class failure

uuuyuqi:
Fix: workaround for Seata autowiring issue by shadowing its upcoming release code locally.

None


LearningGp (approved):
LGTM

PR #4254, opened 2026-02-08
fix: nacos config array/list type refresh not triggered

### LINKED ISSUE #4239: Can not refresh when config using array in yaml
**Which Component**
spring-alibaba-nacos-config:2025.0.0.0

**Describe the bug**

I defined a field of type Set

Image

The YAML config in nacos like this

Image

When I change the value, the value of the field is not refreshed.

Here's why

Image

The key  of ConfigChangeItem is `effectiveNewMenuStoreCodes[3]`, but The interestedKeys is just `effectiveNewMenuStoreCodes`

Zjianru:
Fixes #4239.

## Problem

When YAML configuration contains array/list types (e.g. `Set<String>`, `List<String>`), the Nacos config change parser (`ScaYamlConfigChangeParser`) flattens keys with array indices:

```yaml
storeCodes:
  - code1
  - code2
  - code3
```

Produces `ConfigChangeItem` keys like `storeCodes[0]`, `storeCodes[1]`, `storeCodes[2]`.

However, users register interest in the **base key** via `@NacosConfigKeysListener(interestedKeys = "storeCodes")`.

The existing matching logic uses exact match:

```java
interestedKeys.contains(changeItem.getKey())
// "storeCodes" != "storeCodes[3]" → always false
```

So config changes to array elements **never trigger a refresh**.

## Fix

Added `matchesInterestedKey()` method that:
1. First tries exact match (no regression)
2. If no exact match, strips the first `[index]` suffix and checks the base key

This handles `storeCodes[3]` → `storeCodes`, `config.items[0].name` → `config.items`, etc.

## Tests

Added 6 unit tests: exact match, array-indexed match, nested array match, unrelated key rejection, prefix match, and empty keys passthrough. All pass.

```
Tests run: 6, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```


uuuyuqi (changes requested):
The package path of the test class seems incorrect. It needs to be updated.

Zjianru:
Thanks for catching that! Fixed in the latest commit — moved the test class from the dotted directory (`com.alibaba.cloud.nacos/`) to the proper package path (`com/alibaba/cloud/nacos/annotation/`). Sorry for the careless mistake.

感谢指出！已在最新提交中修复——将测试类从点号目录（`com.alibaba.cloud.nacos/`）移到了正确的包路径（`com/alibaba/cloud/nacos/annotation/`）。抱歉，是我们写的时候不小心放错了位置。

uuuyuqi (approved):
LGTM

PR #4283, opened 2026-03-20
fix: handle NPE when Nacos NotifyCenter.INSTANCE is null during graceful shutdown

Senrian:
## What
When Spring context closes, `NacosGracefulShutdownDelegate` calls `autoServiceRegistration.stop()` which internally deregisters subscribers from Nacos `NotifyCenter`. However, if the Nacos client shutdown hooks have already destroyed `NotifyCenter.INSTANCE`, this results in:

```
java.lang.NullPointerException: Cannot read field "sharePublisher" because "com.alibaba.nacos.common.notify.NotifyCenter.INSTANCE" is null
  at com.alibaba.nacos.common.notify.NotifyCenter.deregisterSubscriber(NotifyCenter.java:242)
  at com.alibaba.nacos.client.naming.remote.gprc.NamingGrpcClientProxy.shutdown(NamingGrpcClientProxy.java:537)
  at com.alibaba.nacos.client.naming.remote.NamingClientProxyDelegate.shutdown(NamingClientProxyDelegate.java:212)
  at com.alibaba.nacos.client.naming.NacosNamingService.shutDown(NacosNamingService.java:637)
  at com.alibaba.cloud.nacos.NacosServiceManager.nacosServiceShutDown(NacosServiceManager.java:117)
  at com.alibaba.cloud.nacos.registry.NacosServiceRegistry.close(NacosServiceRegistry.java:120)
```

## Fix
Wrap the `stop()` call in an inner try-catch to gracefully handle this race condition by logging a warning instead of propagating the error.

## Testing
- Existing `NacosGracefulShutdownDelegateTests` continue to pass

## Related Issues
Fixes #4247
Fixes #4081

PR #3985, opened 2025-05-19
:wrench: use Maven Central Portal

xuxiaowei-com-cn:
https://github.com/alibaba/nacos/issues/13358

settings.xml servers use:

```
<server>
  <id>central</id>
  <username>${CENTRAL_USERNAME}</username>
  <password>${CENTRAL_PASSWORD}</password>
</server>
```

PR #4284, opened 2026-03-22
fix: handle NPE when NotifyCenter.INSTANCE is null during graceful shutdown

Senrian:
## What

When Spring context closes, `NacosGracefulShutdownDelegate.doGracefulShutdown()` calls `autoServiceRegistration.stop()` which internally deregisters subscribers from Nacos `NotifyCenter`. However, if the Nacos client shutdown hooks have already destroyed `NotifyCenter.INSTANCE`, this results in:

```
java.lang.NullPointerException: Cannot read field "sharePublisher" because "NotifyCenter.INSTANCE" is null
```

## Fix

Add explicit `NullPointerException` catch in `doGracefulShutdown()` and log a warning instead of propagating, since this NPE is safe to ignore during shutdown when the NotifyCenter has already been cleaned up.

## Testing

The existing test `sameContextShouldTriggerStop` and other tests in `NacosGracefulShutdownDelegateTests` continue to pass since they mock `autoServiceRegistration.stop()` to not throw NPE.

Fixes #4283

### DIFF
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/main/java/com/alibaba/cloud/nacos/registry/NacosGracefulShutdownDelegate.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/main/java/com/alibaba/cloud/nacos/registry/NacosGracefulShutdownDelegate.java
@@ -2,8 +2,8 @@
  * Copyright 2013-present the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


Senrian:
CLA Status Check: Could a maintainer confirm if the CLA for this PR is complete? I had signed it but the status check is still failing. If there's an issue, please let me know so I can resolve it. The fix addresses a real NPE during graceful shutdown.

xuxiaowei-com-cn:
image
@Senrian

xuxiaowei-com-cn:
https://github.com/alibaba/spring-cloud-alibaba/pull/3985

image

PR #4021, opened 2025-07-15
fix:#4020 The log was incorrectly printed: "A cross-cluster call occurs"

HDHXTM:
### Describe what this PR does / why we need it
Due to this pull request（https://github.com/alibaba/spring-cloud-alibaba/pull/3771）, the default value for com.alibaba.cloud.nacos.NacosDiscoveryProperties#clusterName was removed. When the user does not explicitly specify a clusterName, the method com.alibaba.cloud.nacos.loadbalancer.NacosLoadBalancer#getInstanceResponse will directly enter the else block, causing the message "A cross-cluster call occurs" to be logged incorrectly.

### Does this pull request fix one issue?
Fixes #4020

PR #4287, opened 2026-03-27
fix: incorrect cross-cluster call warning when clusterName is not set

### LINKED ISSUE #4020: The log was incorrectly printed: "A cross-cluster call occurs"
**Which Component**
NacosLoadBalancer

**Describe the bug**
```
String clusterName = this.nacosDiscoveryProperties.getClusterName();
List<ServiceInstance> instancesToChoose = serviceInstances;
if (StringUtils.isNotBlank(clusterName)) {
    List<ServiceInstance> sameClusterInstances = serviceInstances.stream()
                             .filter(serviceInstance -> {
				     String cluster = serviceInstance.getMetadata().get("nacos.cluster");
				     return StringUtils.equals(cluster, clusterName);
				 }).collect(Collectors.toList());
    if (!CollectionUtils.isEmpty(sameClusterInstances)) {
        instancesToChoose = sameClusterInstances;
    }
} else {
    log.warn("A cross-cluster call occurs，name = {}, clusterName = {}, instance = {}",
						serviceId, clusterName, serviceInstances);
}
```
Due to this pull request（https://github.com/alibaba/spring-cloud-alibaba/pull/3771）, the default value for com.alibaba.cloud.nacos.NacosDiscoveryProperties#clusterName was removed. When the user does not explicitly specify a clusterName, the method com.alibaba.cloud.nacos.loadbalancer.NacosLoadBalancer#getInstanceResponse will directly enter the else block, causing the message "A cross-cluster call occurs" to be logged incorrectly.
应修改为：
```
String clusterName = this.nacosDiscoveryProperties.getClusterName();
List<ServiceInstance> instancesToChoose = serviceInstances;
if (StringUtils.isNotBlank(clusterName)) {
    List<ServiceInstance> sameClusterInstances = serviceInstances.stream()
                             .filter(serviceInstance -> {
				     String cluster = serviceInstance.getMetadata().get("nacos.cluster");
				     return StringUtils.equals(cluster, clusterName);
				 }).collect(Collectors.toList());
    if (!CollectionUtils.isEmpty(sameClusterInstances)) {
        instancesToChoose = sameClusterInstances;
    } else {
    log.warn("A cross-cluster call occurs，name = {}, clusterName = {}, instance = {}",
						serviceId, clusterName, serviceInstances);
    }
} 
```
			

**Additional context**
spring-cloud-alibaba-dependencies 2021.0.6.2

daguimu:

In `NacosLoadBalancer.getInstanceResponse()`, the "A cross-cluster call occurs" warning log is incorrectly placed in the outer `else` block, which triggers when `clusterName` is blank (null or empty). Since PR #3771 removed the default value for `clusterName`, this warning fires on every load balancing request when users don't explicitly configure a cluster name.

The warning should only appear when `clusterName` IS configured but no same-cluster instances are found — that is the actual cross-cluster call scenario.

Fixes #4020

### Describe how you did it

Moved the `else` block with the warning log from the outer `if (StringUtils.isNotBlank(clusterName))` into the inner `if (!CollectionUtils.isEmpty(sameClusterInstances))`.

Before (incorrect):
```java
if (StringUtils.isNotBlank(clusterName)) {
    // filter same cluster instances
    if (!CollectionUtils.isEmpty(sameClusterInstances)) {
        instancesToChoose = sameClusterInstances;
    }
}
else {
    log.warn("A cross-cluster call occurs..."); // triggers when clusterName is BLANK
}
```

After (correct):
```java
if (StringUtils.isNotBlank(clusterName)) {
    // filter same cluster instances
    if (!CollectionUtils.isEmpty(sameClusterInstances)) {
        instancesToChoose = sameClusterInstances;
    }
    else {
        log.warn("A cross-cluster call occurs..."); // triggers when clusterName is SET but no match
    }
}
```

### Describe how to verify it

1. Start a Spring Cloud Alibaba application with Nacos discovery without setting `spring.cloud.nacos.discovery.cluster-name`
2. Make load-balanced requests
3. Verify no "A cross-cluster call occurs" warning in logs

### Tests Added

| Change Point | Test |
|-------------|------|
| Blank clusterName should not warn | `blankClusterNameShouldNotWarnCrossCluster()` |
| Empty clusterName should not warn | `emptyClusterNameShouldNotWarnCrossCluster()` |
| Same cluster instances selected correctly | `sameClusterShouldSelectSameClusterInstance()` |
| No same-cluster instances still returns response | `noSameClusterInstancesShouldStillReturnResponse()` |
| Empty instances returns EmptyResponse | `emptyServiceInstancesShouldReturnEmptyResponse()` |

### Special notes for reviews

PR #4021 proposed the same fix for the `2021.x` branch. This PR applies it to the current `2025.1.x` branch with comprehensive unit tests.

### DIFF
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/main/java/com/alibaba/cloud/nacos/loadbalancer/NacosLoadBalancer.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/main/java/com/alibaba/cloud/nacos/loadbalancer/NacosLoadBalancer.java
@@ -60,6 +60,10 @@ public class NacosLoadBalancer implements ReactorServiceInstanceLoadBalancer {
 
 	private static final String IPV4_REGEX = "((2(5[0-5]|[0-4]\\d))|[0-1]?\\d{1,2})(.((2(5[0-5]|[0-4]\\d))|[0-1]?\\d{1,2})){3}";
 
+	private static final long CROSS_CLUSTER_WARN_INTERVAL_MS = 10_000L;
+
+	private volatile long lastWarnLogTime;
+
 	private static final String IPV6_KEY = "IPv6";
 	/**
 	 * Storage local valid IPv6 address, it's a flag whether local machine support IPv6 address stack.
@@ -152,11 +156,9 @@ public class NacosLoadBalancer implements ReactorServiceInstanceLoadBalancer {
 				if (!CollectionUtils.isEmpty(sameClusterInstances)) {
 					instancesToChoose = sameClusterInstances;
 				}
-			}
-			else {
-				log.warn(
-						"A cross-cluster call occurs，name = {}, clusterName = {}, instance = {}",
-						serviceId, clusterName, serviceInstances);
+				else {
+					warnCrossClusterThrottled(clusterName, serviceInstances);
+				}
 			}
 			instancesToChoose = this.filterInstanceByIpType(instancesToChoose);
 
@@ -180,6 +182,15 @@ public class NacosLoadBalancer implements ReactorServiceInstanceLoadBalancer {
 		catch (Exception e) {
 			log.warn("NacosLoadBalancer error", e);
 			return null;
+		}
+	}
+
+	private void warnCrossClusterThrottled(String clusterName, List<ServiceInstance> serviceInstances) {
+		long now = System.currentTimeMillis();
+		if (now - lastWarnLogTime > CROSS_CLUSTER_WARN_INTERVAL_MS) {
+			lastWarnLogTime = now;
+			log.warn("A cross-cluster call occurs，name = {}, clusterName = {}, instance = {}",
+					serviceId, clusterName, serviceInstances);
 		}
 	}
 
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/test/java/com/alibaba/cloud/nacos/loadbalancer/NacosLoadBalancerTests.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/test/java/com/alibaba/cloud/nacos/loadbalancer/NacosLoadBalancerTests.java
@@ -1,0 +1,229 @@
+/*
+ * Copyright 2013-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


uuuyuqi:
Thanks for the fix. One follow-up concern: this warning is still emitted from the request path in `NacosLoadBalancer#getInstanceResponse()`. That means if a service stays in a real cross-cluster fallback state for some time, the log can still be printed once per request and flood the logs under traffic.

Would you consider adding some form of throttling here, for example:
- log at most once per time window per `(serviceId, clusterName)`, or
- log only on state transition (entering cross-cluster fallback), and suppress repeats until recovery

The current fix removes the false positives, which is great, but rate-limiting the warning would make the behavior much safer in sustained fallback scenarios.

daguimu:
@uuuyuqi Good point! I've updated this PR to include log throttling — the cross-cluster warning now logs at most once per 5 minutes per `(serviceId, clusterName)` using a `ConcurrentHashMap<String, Long>` timestamp tracker. This prevents log flooding under sustained cross-cluster fallback traffic while still surfacing the warning when it first occurs.

Added test: `crossClusterWarnShouldBeThrottled()` — verifies the throttle state is tracked correctly after repeated cross-cluster calls.

spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/main/java/com/alibaba/cloud/nacos/loadbalancer/NacosLoadBalancer.java
```diff
@@ -183,4 +187,17 @@ private Response<ServiceInstance> getInstanceResponse(Request<?> request,
 		}
 	}
 
+	private void warnCrossClusterThrottled(String clusterName, List<ServiceInstance> serviceInstances) {
```
uuuyuqi, line 188: Map would also need a maximum keysize limit, but I do not think it needs to be that complex.

We probably do not need to distinguish by `clusterName` or keep a `ConcurrentHashMap` for this warning. A much simpler approach should be enough for this PR: keep a single `lastWarnLogTime` on `NacosLoadBalancer` and only log once every 10 seconds, e.g. `currentMillis - lastWarnLogTime > 10s`.

I do not think we need locking for this field either. This is only best-effort log throttling, so occasional duplicate logs under concurrency are acceptable from an engineering perspective. If we want a minimal concurrency safeguard while keeping it simple, `volatile long lastWarnLogTime` would be sufficient.

uuuyuqi (changes requested):

spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/main/java/com/alibaba/cloud/nacos/loadbalancer/NacosLoadBalancer.java
daguimu: Fixed. Replaced `ConcurrentHashMap` with a single `volatile long lastWarnLogTime`, throttle interval set to 10 seconds, no locking. Updated tests accordingly.

uuuyuqi (approved):
LGTM

Issue #4289, opened 2026-03-28
Spring Cloud Alibaba 2025.1.0.1-SNAPSHOT Seata NoClassDefFoundError UndoLogParserFactory$SingletonHolder


xuxiaowei-com-cn:
# 中文

1. 项目地址：https://github.com/xuxiaowei-com-cn/spring-cloud-alibaba-example
2. 模块：[spring-cloud-2025.1.x/user-seata](https://github.com/xuxiaowei-com-cn/spring-cloud-alibaba-example/tree/main/spring-cloud-2025.1.x/user-seata)
3. 正常代码：https://github.com/xuxiaowei-com-cn/spring-cloud-alibaba-example/tree/3c48b3b6ced0d22bdda4fa8a95eac24be30ef7a1
    1. 流水线运行正常：https://github.com/xuxiaowei-com-cn/spring-cloud-alibaba-example/actions/runs/23678140934
4. 异常代码：https://github.com/xuxiaowei-com-cn/spring-cloud-alibaba-example/tree/4e15f1c727c6f2efaeb1d8a90993faf79929474b
    1. 流水线运行异常：https://github.com/xuxiaowei-com-cn/spring-cloud-alibaba-example/actions/runs/23678228215
    2. 差异变更：https://github.com/xuxiaowei-com-cn/spring-cloud-alibaba-example/commit/4e15f1c727c6f2efaeb1d8a90993faf79929474b
5. 复现流程：
    1. 准备数据库：`sql` 文件夹
    2. 启动 `nacos-bootstrap-3.x` Maven 模块
    3. 启动 `seata-server-2.x` Maven 模块
    4. 启动 `spring-cloud-2025.1.x/user-seata` Maven 模块
    5. 运行 `spring-cloud-2025.1.x/user-seata/src/test/java/cn/com/xuxiaowei/user/controller/SeataRestController_2025_1_x_Tests.java` 测试类，断言失败
    6. 查看 `spring-cloud-2025.1.x/user-seata` 日志，异常内容见下方

# English

1. Project address: https://github.com/xuxiaowei-com-cn/spring-cloud-alibaba-example
2. Module: [spring-cloud-2025.1.x/user-seata](https://github.com/xuxiaowei-com-cn/spring-cloud-alibaba-example/tree/main/spring-cloud-2025.1.x/user-seata)
[...]

PR #4290, opened 2026-03-29
:heavy_plus_sign: add the dependency jackson-datatype-jsr310 required by JacksonUndoLogParser of seata 2.6.0

xuxiaowei-com-cn:
add the dependency com.fasterxml.jackson.datatype:jackson-datatype-jsr310 required by JacksonUndoLogParser of seata 2.6.0

https://github.com/alibaba/spring-cloud-alibaba/issues/4289

https://github.com/alibaba/spring-cloud-alibaba/issues/4289

### DIFF
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-seata/pom.xml
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-seata/pom.xml
@@ -55,6 +55,11 @@
             <artifactId>seata-spring-boot-starter</artifactId>
         </dependency>
 
+		<dependency>
+			<groupId>com.fasterxml.jackson.datatype</groupId>
+			<artifactId>jackson-datatype-jsr310</artifactId>
+		</dependency>
+
         <dependency>
             <groupId>org.springframework.cloud</groupId>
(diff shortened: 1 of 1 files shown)


xuxiaowei-com-cn:
- PR：https://github.com/apache/incubator-seata/pull/7908

uuuyuqi:
I think this PR is reasonable as a short-term compatibility fix and can be merged first to keep `2025.1.x` usable.

From the current investigation, `2025.1.0.1-SNAPSHOT` introduced a regression after upgrading Seata from `2.5.0` to `2.6.0`, and `#4290` is a minimal SCA-side fix that restores the missing runtime dependency required by the Seata undo log parser path.

I would treat this as a temporary integration patch rather than the final long-term solution. The long-term fix should still come from consuming an upstream Seata release that includes the relevant fix. After SCA upgrades to that Seata version, we should re-verify this scenario and remove the extra dependency if it is no longer necessary.

So my suggestion is:
- merge `#4290` now to avoid leaving the current snapshot broken
- track the upstream Seata release
- clean up this extra dependency only after upgrading Seata and confirming the issue is resolved without it

uuuyuqi (approved):
LGTM

PR #4311, opened 2026-04-23
fix: allow sharing RocketMQ MQClientInstance to avoid thread explosion

### LINKED ISSUE #4291: Consumers with different groups create separate MQClientInstance per binding, causing thread explosion
When using spring-cloud-starter-stream-rocketmq with many consumer bindings (e.g. 70+), each consumer creates its own MQClientInstance, resulting in massive thread overhead. In our production service with 71 consumer bindings, the JVM spawned ~1,700 threads solely from RocketMQ (total 2,200+), most of which   
  are redundant Netty I/O, Rebalance, PullMessage, and Heartbeat threads.          
                                                                                                                                                                                                                                                                                                                         
  Root Cause                                                                       
                                                                                                                                                                                                                                                                                                                         
  RocketMQConsumerFactory.initPushConsumer() calls RocketMQUtils.getInstanceName() to set each consumer's instanceName:                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                         
  // RocketMQConsumerFactory.java line 77-78                                                                                                                                                                                                                                                                             
  consumer.setInstanceName(                                                                                                                                                                                                                                                                                              
      RocketMQUtils.getInstanceName(rpcHook, consumerProperties.getGroup()));                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                         
  RocketMQUtils.getInstanceName() appends System.nanoTime() to ensure uniqueness:                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                         
  // RocketMQUtils.java line 85-86                                                                                                                                                                                                                                                                                       
  instanceName.append(identify).append(separator).append(UtilAll.getPid())                                                                                                                                                                                                                                               
          .append(separator).append(Long.toString(System.nanoTime(), 36));                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                         
  Since each consumer gets a unique instanceName, they each generate a unique clientId (IP@instanceName), and MQClientManager.getOrCreateMQClientInstance() creates a separate MQClientInstance for each one. Every MQClientInstance spawns ~20-24 threads (Netty EventLoop, PullMessageService, RebalanceService,       
  ScheduledExecutor, etc.).                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                         
  Why this is unnecessary                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                         
  RocketMQ's MQClientInstance.consumerTable uses consumer group name as the key. Consumers with different groups connecting to the same NameServer can safely share a single MQClientInstance — this is RocketMQ's native supported behavior. The unique instanceName per consumer is unnecessary when groups are already
   distinct.                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                         
  Impact                                                                                                                                                                                                                                                                                                                 
                                                                                   
  ┌────────────────────────┬────────────────────────┐                                                                                                                                                                                                                                                                    
  │         Metric         │  Actual (71 bindings)  │                                                                                                                                                                                                                                                                    
  ├────────────────────────┼────────────────────────┤                                                                                                                                                                                                                                                                    
  │ MQClientInstance count │ 72 (should be 1)       │                                                                                                                                                                                                                                                                    
  ├────────────────────────┼────────────────────────┤                              
  │ RocketMQ threads       │ ~1,700 (should be ~24) │                                                                                                                                                                                                                                                                    
  ├────────────────────────┼────────────────────────┤                              
  │ Total JVM threads      │ 2,200+                 │                
  └────────────────────────┴────────────────────────┘                                                                                                                                                                                                                                                                    
                                                                    
  Expected Behavior                                                                                                                                                                                                                                                                                                      
                                                                                   
  Consumers connecting to the same NameServer should share a single MQClientInstance by default, unless explicitly configured otherwise.
                                                                                                                                                                                                                                                                                                                         
  Suggested Fix                                     
                                                                                                                                                                                                                                                                                                                         
  Provide a configurable instanceName strategy. For example:                       
                                                                  
  Option A — Use a fixed instanceName by default (group name uniqueness already prevents conflicts):                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                         
  // RocketMQConsumerFactory.java                                                                                                                                                                                                                                                                                        
  consumer.setInstanceName(                                                                                                                                                                                                                                                                                              
      RocketMQUtils.getInstanceName(rpcHook, consumerProperties.getGroup()));      
  // change to:                                    
  consumer.setInstanceName(                                          
      RocketMQUtils.getSharedInstanceName(rpcHook));                                                                                                                                                                                                                                                                     
                                                                  
  Option B — Expose a configuration property:                                                                                                                                                                                                                                                                            
                                                                                   
  spring:                                                                                                                                                                                                                                                                                                                
    cloud:                                                                         
      stream:                                                                                                                                                                                                                                                                                                            
        rocketmq:                                                 
          binder:                                                                                                                                                                                                                                                                                                        
            share-client-instance: true  # default true                            
                                                   
  Option C — At minimum, provide a hook (e.g. ConsumerEndpointCustomizer timing fix or a ClientInstanceCustomizer callback) so users can override the instanceName after pushConsumer is created but before start() is called. Currently there is no such hook — ConsumerEndpointCustomizer.configure() is called before 
  afterPropertiesSet(), at which point pushConsumer is still null.   
                                                                                                                                                                                                                                                                                                                         
  Why users cannot work around this                                                                                                                                                                                                                                                                                      
                                                                     
  The framework's lifecycle leaves no clean extension point:                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                         
  createConsumerEndpoint()       → new Adapter()         [pushConsumer = null]
  ConsumerEndpointCustomizer     → configure()           [pushConsumer = null, can't modify]                                                                                                                                                                                                                             
  afterPropertiesSet()           → onInit() → Factory    [pushConsumer created with unique instanceName]
  start()                        → pushConsumer.start()  [MQClientInstance created, too late]
                                                                     
  - BeanPostProcessor — doesn't work because RocketMQInboundChannelAdapter is never registered as a Spring bean (created via new in doBindConsumer)                                                                                                                                                                      
  - ConsumerEndpointCustomizer — called before pushConsumer exists                                                                                                                                                                                                                                                       
  - BinderCustomizer — can access the binder but not the individual consumers                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                         
  The only workaround is classpath shadowing of RocketMQUtils, which is fragile and not maintainable.                                                                                                                                                                                                                    
                                                                     
  Environment                                                                                                                                                                                                                                                                                                            
                                                                                   
  - spring-cloud-starter-stream-rocketmq: 2023.0.3.2                                                                                                                                                                                                                                                                     
  - rocketmq-client: 5.3.1                                                                                                                                                                                                                                                                                               
  - Spring Boot: 3.x                                                                                                                                                                                                                                                                                                     
  - Java: 17

daguimu:
## Problem

`RocketMQUtils.getInstanceName` appends `System.nanoTime()` to every generated client name, so every consumer and producer ends up with a unique `instanceName` and therefore a unique `clientId` (`ip@instanceName`). RocketMQ's `MQClientManager.getOrCreateMQClientInstance` creates a separate `MQClientInstance` per unique `clientId`, and each instance spawns ~20-24 threads (Netty EventLoop, PullMessageService, RebalanceService, ScheduledExecutor, etc.). With many stream bindings (e.g. 70+ consumers in the reported case) this multiplies into ~1,700 RocketMQ threads where one shared instance would be sufficient.

## Root Cause

Consumers with different groups connecting to the same name server can safely share a single `MQClientInstance` — the `consumerTable` inside `MQClientInstance` is keyed by group. The `nanoTime` suffix defeats that sharing unconditionally, even when the binding topology would tolerate it.

## Fix

- Add an opt-in `shareClientInstance` boolean to `RocketMQCommonProperties` (default `false`, so behavior is unchanged unless the user sets it).
- Add `RocketMQUtils.getInstanceName(rpcHook, identify, shareClientInstance)`. When `shareClientInstance` is `true`, the `nanoTime` segment is omitted and the generated name becomes deterministic (`[ak|]identify|pid`), which lets RocketMQ reuse a single `MQClientInstance` across multiple consumers/producers in the same JVM.
- Keep the existing two-argument `getInstanceName` as a thin delegate so any external callers and the existing default behavior are preserved.
- Propagate `shareClientInstance` from `RocketMQBinderConfigurationProperties` to the consumer/producer properties inside `mergeRocketMQProperties` using OR-logic, so users can toggle the flag once at the binder level.
- Route the flag through `RocketMQConsumerFactory#initPushConsumer`, `RocketMQConsumerFactory#initPullConsumer`, and `RocketMQProduceFactory` so every produced/consumed client honors it.

Users who hit the thread explosion can now set:

```yaml
spring:
  cloud:
    stream:
      rocketmq:
        binder:
          share-client-instance: true
```

or scope it per-binding on consumer/producer properties.

## Tests Added

New unit tests in `RocketMQUtilsTest`:

| Change Point | Test |
|-------------|------|
| Default overload keeps pre-existing per-call uniqueness | `getInstanceNameWithoutSharingProducesDistinctNames`, `getInstanceNameWithoutSharingIncludesNanoTimeSuffix` |
| New `shareClientInstance=true` mode yields deterministic names | `getInstanceNameWithSharingProducesDeterministicNameForSameIdentify`, `getInstanceNameWithSharingOmitsNanoTimeSuffix` |
| Shared names still differentiate by `identify` | `getInstanceNameWithSharingStillDifferentiatesByIdentify` |
| Access-key segment still prepended when an `AclClientRPCHook` is supplied | `getInstanceNameWithRpcHookPrependsAccessKey` |
| Two-arg overload delegates to the unshared path | `defaultOverloadDelegatesToUnsharedBehavior` |
| `mergeRocketMQProperties` OR-logic for the new flag | `mergeRocketMQPropertiesPropagatesShareClientInstanceFromBinder`, `mergeRocketMQPropertiesPreservesConsumerShareClientInstanceWhenBinderIsFalse`, `mergeRocketMQPropertiesDefaultLeavesShareClientInstanceFalse` |

## Impact

- Default behavior is unchanged (`shareClientInstance=false`), so existing applications pick up no behavioral drift.
- Applications with many RocketMQ stream bindings can opt in to collapse per-binding `MQClientInstance` copies into one, recovering the ~20 threads each duplicate would spawn.
- Changes are confined to the RocketMQ stream binder starter; no cross-module surface is altered.

Fixes #4291

### DIFF
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-stream-rocketmq/src/main/java/com/alibaba/cloud/stream/binder/rocketmq/integration/inbound/RocketMQConsumerFactory.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-stream-rocketmq/src/main/java/com/alibaba/cloud/stream/binder/rocketmq/integration/inbound/RocketMQConsumerFactory.java
@@ -81,7 +81,7 @@ public final class RocketMQConsumerFactory {
 				null == rpcHook && consumerProperties.getVipChannelEnabled());
 		consumer.setInstanceName(
 				RocketMQUtils.getInstanceName(actualRpcHook, group,
-						consumerProperties.isShareClientInstance()));
+						Boolean.TRUE.equals(consumerProperties.getShareClientInstance())));
 		consumer.setNamespace(consumerProperties.getNamespace());
 		consumer.setNamespaceV2(consumerProperties.getNamespaceV2());
 		consumer.setNamesrvAddr(consumerProperties.getNameServer());
@@ -153,7 +153,7 @@ public final class RocketMQConsumerFactory {
 				null == rpcHook && consumerProperties.getVipChannelEnabled());
 		consumer.setInstanceName(
 				RocketMQUtils.getInstanceName(actualRpcHook, consumerGroup,
-						consumerProperties.isShareClientInstance()));
+						Boolean.TRUE.equals(consumerProperties.getShareClientInstance())));
 		if (null != allocateMessageQueueStrategy) {
 			consumer.setAllocateMessageQueueStrategy(allocateMessageQueueStrategy);
 		}
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-stream-rocketmq/src/main/java/com/alibaba/cloud/stream/binder/rocketmq/integration/inbound/RocketMQConsumerFactory.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-stream-rocketmq/src/main/java/com/alibaba/cloud/stream/binder/rocketmq/integration/inbound/RocketMQConsumerFactory.java
@@ -80,7 +80,8 @@ public final class RocketMQConsumerFactory {
 		consumer.setVipChannelEnabled(
 				null == rpcHook && consumerProperties.getVipChannelEnabled());
 		consumer.setInstanceName(
-				RocketMQUtils.getInstanceName(actualRpcHook, group));
+				RocketMQUtils.getInstanceName(actualRpcHook, group,
+						consumerProperties.isShareClientInstance()));
 		consumer.setNamespace(consumerProperties.getNamespace());
 		consumer.setNamespaceV2(consumerProperties.getNamespaceV2());
 		consumer.setNamesrvAddr(consumerProperties.getNameServer());
@@ -151,7 +152,8 @@ public final class RocketMQConsumerFactory {
 		consumer.setVipChannelEnabled(
 				null == rpcHook && consumerProperties.getVipChannelEnabled());
 		consumer.setInstanceName(
-				RocketMQUtils.getInstanceName(actualRpcHook, consumerGroup));
+				RocketMQUtils.getInstanceName(actualRpcHook, consumerGroup,
+						consumerProperties.isShareClientInstance()));
 		if (null != allocateMessageQueueStrategy) {
 			consumer.setAllocateMessageQueueStrategy(allocateMessageQueueStrategy);
 		}
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-stream-rocketmq/src/main/java/com/alibaba/cloud/stream/binder/rocketmq/integration/outbound/RocketMQProduceFactory.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-stream-rocketmq/src/main/java/com/alibaba/cloud/stream/binder/rocketmq/integration/outbound/RocketMQProduceFactory.java
@@ -114,7 +114,8 @@ public final class RocketMQProduceFactory {
 				null == rpcHook && producerProperties.getVipChannelEnabled());
 		producer.setInstanceName(
 				RocketMQUtils.getInstanceName(actualRpcHook,
-						topic + "|" + UtilAll.getPid()));
+						topic + "|" + UtilAll.getPid(),
+						producerProperties.isShareClientInstance()));
 		producer.setNamesrvAddr(producerProperties.getNameServer());
 		producer.setNamespaceV2(producerProperties.getNamespaceV2());
 		producer.setSendMsgTimeout(producerProperties.getSendMsgTimeout());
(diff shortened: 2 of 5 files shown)


oss-sentinel-ai (approved):
﻿### Summary
Well-designed performance fix. The root cause analysis is thorough: each consumer/producer gets a unique `instanceName` (via `nanoTime` suffix), causing RocketMQ to create separate `MQClientInstance` per client (~20-24 threads each). With 70+ consumers this explodes to ~1700 threads. The opt-in `shareClientInstance` flag correctly addresses this by omitting the `nanoTime` suffix, allowing RocketMQ's built-in client sharing. Backward-compatible (default `false`), supports per-binding override, and has comprehensive test coverage (12 tests).

LGTM — no issues found.

---

Issue #4296, opened 2026-04-07
Question: Is the fix in #3644 effective in version 2025.1.0.0?


qnnn:
Hi, I have a question about this fix: #3644

While checking the code in 2025.1.0.0, it seems the logic introduced by this fix may not be triggered. I'm not sure if I missed something.

Could you confirm whether this fix is included and expected to work in 2025.1.0.0?

Thanks!

**Which Component**
Nacos Config

uuuyuqi:
I checked the 2025.1.0.0 code path again, and I think this concern is valid.

The fix from #3644 is included in 2025.1.0.0. `NacosContextRefresher` stores the pushed config into `NacosSnapshotConfigManager`, and `NacosPropertySourceBuilder` can consume that snapshot before falling back to `configService.getConfig(...)`.

However, in the default ConfigData refresh path, the reload does not appear to go through `NacosPropertySourceBuilder`. The Nacos refresh event is converted to a Spring Cloud `RefreshEvent`, and Spring Cloud refreshes the environment through `ConfigDataContextRefresher`. That path reloads Nacos config via `NacosConfigDataLoader`.

The problem is that `NacosConfigDataLoader#pullConfig(...)` directly calls `configService.getConfig(dataId, group, timeout)`. It does not read from `NacosSnapshotConfigManager`, and it does not go through `NacosPropertySourceBuilder`.

So although #3644 is merged, its snapshot protection can be bypassed in the default ConfigData refresh flow in 2025.1.0.0. This means the original stale-read problem in a Nacos cluster can still happen in that path.

This is separate from #4312, which fixes file extension preservation during refresh. It does not address whether the pushed config snapshot from #3644 is consumed by the ConfigData reload path. @daguimu

DragonFSKY:
I’d like to contribute a fix for the ConfigData reload path discussed here.

My understanding is that `NacosContextRefresher` already stores the pushed config content in `NacosSnapshotConfigManager`, but the default ConfigData refresh flow reloads config through `NacosConfigDataLoader`, which currently falls back directly to `ConfigService#getConfig(...)`.

My proposed approach is to let `NacosConfigDataLoader` first consume `NacosSnapshotConfigManager`, and only fall back to `ConfigService#getConfig(...)` when no valid snapshot is available. This keeps the existing fallback behavior while applying the snapshot protection to the ConfigData path as well.

I opened PR #4319 with this change. Please let me know if this direction looks reasonable or if there are any concerns.

[...]

PR #4312, opened 2026-04-23
fix: preserve Nacos config file extension across refresh

daguimu:
## Problem

When a Nacos config change is pushed to the client, `NacosPropertySourceRefreshListener#handle(NacosConfigRefreshEvent)` rebuilds the affected `NacosPropertySource` by calling:

```java
nacosPropertySourceBuilder.build(dataId, group, "properties", prevpropertySource.isRefreshable());
```

The third argument — the file extension — is hard-coded as `"properties"`. `NacosDataParserHandler#parseNacosData` then selects `PropertiesPropertySourceLoader` regardless of the user's actual format (yaml, yml, json, xml). So after the first refresh, a yaml config like

```yaml
server:
  port: 8080
```

is re-parsed as a properties file and produces a single bogus key `server:\n  port` with value `8080`, replacing the correctly-parsed yaml source that was installed at startup.

## Root Cause

`NacosPropertySource` never remembers the extension it was created with. At refresh time the listener therefore has no way to recover the original format and defaults to `"properties"`. This affects both the main `spring.cloud.nacos.config.file-extension` as well as `extension-configs` / `shared-configs` that may declare their own suffix per entry.

## Fix

- Add an optional `fileExtension` field to `NacosPropertySource` with a new public constructor overload and a `getFileExtension()` accessor. The existing constructors stay for backward compatibility.
- `NacosPropertySourceBuilder#build` already receives the file extension from its callers, so it now passes it into the new constructor.
- `NacosConfigDataLoader#doLoad` propagates `NacosItemConfig#getSuffix()`, which is already the per-entry suffix, so `extension-configs`/`shared-configs` with non-default formats are also preserved.
- `NacosPropertySourceRefreshListener#handle` reads the preserved extension from the previous `NacosPropertySource`, falls back to `NacosConfigProperties#getFileExtension()` (useful when the source was built via the legacy constructor from third-party code), and finally keeps the historical `"properties"` default so nothing regresses.

## Tests Added

| Change Point | Test |
|-------------|------|
| Legacy constructor leaves `fileExtension` unset | `NacosPropertySourceTest#fileExtensionIsNullWhenUsingLegacyConstructor` |
| New constructor preserves the provided extension | `NacosPropertySourceTest#fileExtensionIsPreservedWhenProvided`, `otherPropertiesAreNotAffectedByNewConstructor` |
| New constructor accepts `null` without side effects | `NacosPropertySourceTest#fileExtensionCanBeNullExplicitly` |
| `Builder#build` propagates the yaml extension into the resulting `NacosPropertySource` | `NacosPropertySourceBuilderTest#builtPropertySourcePreservesFileExtension` |
| `Builder#build` propagates the properties extension | `NacosPropertySourceBuilderTest#builtPropertySourcePreservesPropertiesExtension` |

## Impact

- Yaml / yml / json / xml Nacos configs retain their correct parser across pushes — the refresh path now mirrors the initial-load path.
- Properties users see no behavioral change: their extension is preserved explicitly instead of implicitly re-defaulting.
- Third-party code that still calls the legacy `NacosPropertySource` constructor keeps working; the refresh listener's two-step fallback covers that case.
- Scope is limited to `spring-alibaba-nacos-config`; no public contract on `NacosPropertySource` was removed.

Addresses #4296

### DIFF
--- a/spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/client/NacosPropertySource.java
+++ b/spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/client/NacosPropertySource.java
@@ -60,30 +60,39 @@ public class NacosPropertySource extends MapPropertySource {
 	 * File extension this property source was parsed with (e.g. {@code yaml},
 	 * {@code properties}, {@code json}). Captured so that subsequent refreshes
 	 * can re-parse updated config with the same loader instead of falling back
-	 * to the properties default.
+	 * to the properties default. The value is the same notion the rest of the
+	 * codebase tracks as {@code NacosConfigProperties#fileExtension} and
+	 * {@code NacosItemConfig#suffix}.
 	 */
-	private @Nullable String fileExtension;
+	private final @Nullable String fileExtension;
 
 	NacosPropertySource(String group, String dataId, Map<String, Object> source,
 			Date timestamp, boolean isRefreshable) {
+		this(group, dataId, source, timestamp, isRefreshable, null);
+	}
+
+	private NacosPropertySource(String group, String dataId, Map<String, Object> source,
+			Date timestamp, boolean isRefreshable,
+			@Nullable String fileExtension) {
 		super(String.join(NacosConfigProperties.COMMAS, dataId, group), source);
 		this.group = group;
 		this.dataId = dataId;
 		this.timestamp = timestamp;
 		this.isRefreshable = isRefreshable;
+		this.fileExtension = fileExtension;
 	}
 
 	public NacosPropertySource(List<PropertySource<?>> propertySources, String group,
 			String dataId, Date timestamp, boolean isRefreshable) {
 		this(group, dataId, getSourceMap(group, dataId, propertySources), timestamp,
-				isRefreshable);
+				isRefreshable, null);
 	}
 
 	public NacosPropertySource(List<PropertySource<?>> propertySources, String group,
 			String dataId, Date timestamp, boolean isRefreshable,
 			@Nullable String fileExtension) {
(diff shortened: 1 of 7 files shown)


oss-sentinel-ai (approved):
﻿### Summary
This PR fixes the same hardcoded `"properties"` suffix bug in the config refresh path. The approach is clean: immutable `final @Nullable String fileExtension` field with constructor-based initialization, legacy constructor overload for backward compatibility, and a well-documented fallback with diagnostic logging in the refresh listener. Good test coverage.

### Findings
- **[Warning]** This PR, #4349, and #4335 all fix the same hardcoded-suffix bug with different approaches. Only one should be merged. Among the three, this PR's approach (immutable `final` field + constructor overload + diagnostic logging fallback) is the cleanest design. Please coordinate with maintainers to select one.
- **[Info]** The `fileExtension` naming is more descriptive than `suffix` (used in #4349/#4335) — good choice.

LGTM on the implementation quality.

---

oss-sentinel-ai:
⚠️ **Merge conflict detected**

This PR has conflicts with the `2025.1.x` branch and cannot be merged. Please rebase or merge `2025.1.x` into your branch and resolve the conflicts:

```bash
git fetch origin
git checkout fix/nacos-config-refresh-file-extension-issue4296
git rebase origin/2025.1.x
# resolve conflicts, then:
git push --force-with-lease
```

This is a one-time reminder. Feel free to @mention me for a re-review after conflicts are resolved.

---
*Automated notification by github-manager-bot*

uuuyuqi:
Thanks for the contribution. This change has been superseded by #4349, together with #4341 for ConfigData PropertySource exposure. The current 2025.1.x branch preserves the original suffix across consecutive refreshes and includes regression coverage. Closing this PR as superseded.

Note that #4296 tracks a separate snapshot-consumption issue and is not resolved by this PR.

PR #4314, opened 2026-04-25
fix: use null-safe equals and wire recovery callback

srpatcha:
## Summary

Fixes two bugs:

1. **NPE in JobSyncService** — `.equals()` called on nullable fields, causing NullPointerException. Replaced with `Objects.equals()`.
2. **Dead recovery callback** — `recoveryCallback` in RocketMQInboundChannelAdapter was defined but never passed to `retryTemplate.execute()`. Wired it as the second argument.

Also fixes README whitespace.

Signed-off-by: Srikanth Patchava <spatchava@meta.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Spring Cloud uses Maven for most build-related activities, and you should be abl
 
 ## How to Use
 
-### Add maven dependency 
+### Add maven dependency
 
 #### Release Version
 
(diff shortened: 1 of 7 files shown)


srpatcha:
Closing this PR — while attempting to rebase onto the current `2025.1.x` tip, I discovered that upstream has **already converged on the same fix** for the recovery-callback issue this PR was addressing.

Specifically in `RocketMQInboundChannelAdapter.java`, the upstream branch now contains:
```java
this.recoveryCallback.recover(
    ErrorMessageUtils.getAttributeAccessor(message, message),
    cause);
}
else {
    throw retryException;
}
```
which directly fixes the recovery-callback wiring my PR was attempting (just via a different code shape than the `retryTemplate.execute(..., recoveryCallback)` style I used).

Since the central fix is now redundant, and the other bundled commits (README whitespace, `AggregatedHealthIndicator` + tests, `NacosContextRefresher` null checks) are tangentially scoped to this PR's stated purpose, I'm closing here. If any of those side-commits look useful, happy to open them as separate focused PRs — let me know.

Thanks for keeping the project healthy! 🎉

PR #4315, opened 2026-04-29
fix: invoke recovery callback when RocketMQ inbound retries are exhausted

daguimu:
## Problem

`RocketMQInboundChannelAdapter` advertises a `RecoveryCallback` slot, and `RocketMQMessageChannelBinder` wires `errorInfrastructure.getRecoverer()` into it whenever `maxAttempts > 1`. The recoverer is the bridge that turns retry-exhausted messages into events on the binder's error channel (DLQ, retry-exhausted handlers, etc.). After commit 84ad287e (JSpecify annotations) the recovery wiring stopped firing: `consumeMessage` no longer calls the callback. Instead the `RetryException` raised by the retry template is swallowed by the outer per-message catch, which logs and returns the failure supplier so RocketMQ redelivers the same message indefinitely. From the user's perspective the configured DLQ infrastructure is silently inert.

## Root cause

The retry block looks like:

```java
this.retryTemplate.setRetryListener(new RetryListener() {
    @Override
    public void onRetryPolicyExhaustion(...) {
        log.warn("retry exhausted for messageExt: {}", messageExt, exception);
    }
});
this.retryTemplate.execute(...);
```

The exhaustion path only logs. The downstream `RecoveryCallback` is never reached, so anything the binder set up around it is unreachable.

## Fix

Catch `RetryException` from `retryTemplate.execute(...)` directly. When a `recoveryCallback` is configured, hand the underlying cause to it via `ErrorMessageUtils.getAttributeAccessor(message, message)` and let the loop continue, so the success supplier is returned and the message is acknowledged after the configured recovery (typically: published to the error channel that backs the DLQ). When no callback is configured the exception is rethrown, preserving the existing "redeliver via RocketMQ" behavior.

## Tests added

`RocketMQInboundChannelAdapterRecoveryTests` covers both branches by reflectively driving `consumeMessage` against an output channel that always fails:

- `recoveryCallbackInvokedWhenRetriesExhausted` — verifies the callback is invoked exactly once with the underlying cause and that the success supplier is returned.
- `retriesExhaustedWithoutRecoveryFallsBackToFailure` — verifies the previous behavior is preserved when no callback is configured.

Both tests use a zero-delay `FixedBackOff` so they run in a few hundred milliseconds.

### DIFF
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-stream-rocketmq/src/main/java/com/alibaba/cloud/stream/binder/rocketmq/integration/inbound/RocketMQInboundChannelAdapter.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-stream-rocketmq/src/main/java/com/alibaba/cloud/stream/binder/rocketmq/integration/inbound/RocketMQInboundChannelAdapter.java
@@ -36,7 +36,6 @@ import org.slf4j.LoggerFactory;
 
 import org.springframework.cloud.stream.binder.ExtendedConsumerProperties;
 import org.springframework.core.retry.RetryException;
-import org.springframework.core.retry.RetryListener;
 import org.springframework.core.retry.RetryTemplate;
 import org.springframework.integration.context.OrderlyShutdownCapable;
 import org.springframework.integration.core.RecoveryCallback;
@@ -86,8 +85,6 @@ public class RocketMQInboundChannelAdapter extends MessageProducerSupport
 						"Cannot have an 'errorChannel' property when a 'RetryTemplate' is "
 								+ "provided; use an 'ErrorMessageSendingRecoverer' in the 'recoveryCallback' property to "
 								+ "send an error message when retries are exhausted");
-				this.retryTemplate.setRetryListener(new RetryListener() {
-				});
 			}
 			pushConsumer = RocketMQConsumerFactory
 					.initPushConsumer(extendedConsumerProperties);
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-stream-rocketmq/src/main/java/com/alibaba/cloud/stream/binder/rocketmq/integration/inbound/RocketMQInboundChannelAdapter.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-stream-rocketmq/src/main/java/com/alibaba/cloud/stream/binder/rocketmq/integration/inbound/RocketMQInboundChannelAdapter.java
@@ -37,12 +37,11 @@ import org.slf4j.LoggerFactory;
 import org.springframework.cloud.stream.binder.ExtendedConsumerProperties;
 import org.springframework.core.retry.RetryException;
 import org.springframework.core.retry.RetryListener;
-import org.springframework.core.retry.RetryPolicy;
 import org.springframework.core.retry.RetryTemplate;
-import org.springframework.core.retry.Retryable;
 import org.springframework.integration.context.OrderlyShutdownCapable;
 import org.springframework.integration.core.RecoveryCallback;
 import org.springframework.integration.endpoint.MessageProducerSupport;
+import org.springframework.integration.support.ErrorMessageUtils;
 import org.springframework.integration.support.MessageBuilder;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessagingException;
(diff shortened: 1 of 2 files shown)


uuuyuqi:
Nice fix. Reproduced locally with RocketMQ 5.3 + Spring Boot 4: in the buggy version the DLQ handler fires 0 times and the broker keeps redelivering; after applying the patch the DLQ handler fires once and the broker stops redelivering.

A few non-blocking suggestions:

### 1. Consider a test for "recoveryCallback throws"

The new tests cover the two main branches (callback / no callback). When `recoveryCallback.recover(...)` itself throws (e.g. a downstream errorChannel handler fails), behavior falls through to the outer `catch (Exception e)` and returns `failSupplier`. Worth pinning down in a test since it's an implicit contract.

### 2. Lost "retry exhausted" log on the no-callback path

Previously `RetryListener.onRetryPolicyExhaustion` always emitted `retry exhausted for messageExt: ...`. After this PR, that log only fires when `recoveryCallback != null`. The no-callback path now only logs the generic `consume message failed.` from the outer catch, making it harder for operators to distinguish "retries exhausted" from "ordinary consume failure". Suggest adding a dedicated WARN before `throw retryException;`, or hoisting the WARN to the top of the catch block.

### 3. Batch consumption edge case (`consumeMessageBatchMaxSize > 1`)

`consumeMessage` processes the list in a loop; if `recover()` throws for any message, the outer catch returns `failSupplier`, causing the broker to redeliver the **entire** batch — already-recovered messages will be recovered (and pushed to the DLQ) again. Default `batchSize=1` is unaffected, but users who tuned it up could hit this. A proper fix would require per-message ack tracking inside the batch (out of scope here); a comment above the loop, or a doc note recommending `batchSize=1`, would suffice.

### Behavior-change note for release notes

Pre-fix, the broker's `%DLQ%consumerGroup` topic acted as a backstop after 16 reconsume attempts. Post-fix, that path is no longer triggered. Anyone monitoring the broker DLQ topic for alerting will need to switch to the app-side errorChannel. Worth calling out in the release notes.

uuuyuqi (changes requested):
Please review my comment above.

daguimu:
@uuuyuqi pushed 782e13d8 covering all three suggestions:

1. **Lost "retry exhausted" log on the no-callback path** — hoisted the WARN out of the `recoveryCallback != null` branch so it now fires whenever retries are exhausted, regardless of callback. The no-callback path now emits both the dedicated WARN and the existing generic consume-failure WARN, so operators can again distinguish retry exhaustion from ordinary consume failures.

2. **Batch consumption edge case (`consumeMessageBatchMaxSize > 1`)** — added a note above the per-message loop documenting the duplicate-recovery hazard (recover() throwing midway through a batch causes the whole batch to be redelivered and previously-recovered messages re-recovered) and recommending `batchSize=1` with a `recoveryCallback`.

3. **Test for "recoveryCallback throws"** — added `recoveryCallbackThrowingFallsBackToFailureSupplier`, which configures a `RecoveryCallback` that always throws and asserts the outer catch returns the failure supplier so RocketMQ redelivers. Pins down the implicit contract you flagged.

Re the **behavior-change note for release notes** — agreed; the broker `%DLQ%consumerGroup` backstop no longer triggers once this lands and operators monitoring it must switch to the app-side errorChannel. I'll make sure that's called out in the release notes for this fix.

Ready for re-review when you have time.

uuuyuqi (approved):
LGTM

daguimu:
Hi @uuuyuqi — following up on the behavior-change note I promised in my earlier reply. Now that this is merged, here's a suggested wording for the 2025.1.x release notes whenever the next version cuts:

---

**Behavior change — RocketMQ inbound retry-exhausted DLQ pathway**

Before this fix, when a `spring-cloud-stream-binder-rocketmq` consumer exhausted its in-app retries, the configured `RecoveryCallback` (DLQ infrastructure / error channel) was silently bypassed — the `RetryException` was swallowed and RocketMQ kept redelivering the message, until the broker's `%DLQ%<consumerGroup>` topic eventually caught it after the default 16 reconsume attempts.

Starting with this release, retry-exhausted messages are correctly routed through the configured `RecoveryCallback` (typically the binder's `errorChannel` and any DLQ infrastructure subscribed to it).

**Migration note for operators**: if you were monitoring the broker's `%DLQ%<consumerGroup>` topic as your last-resort alerting signal, switch your alerts to the application-side `errorChannel` (or the DLQ topic configured by `errorInfrastructure`). The broker-side `%DLQ%` backstop is no longer reached on the normal retry-exhaustion path.

Refs: #4315

---

Happy to PR this somewhere if there's a release-notes draft or branch you'd like me to target — I didn't see a `CHANGELOG.md` in the repo, so I assume notes live in the GitHub Release body where I can't write directly. Let me know if you'd prefer a different format / location.

PR #4319, opened 2026-05-03
fix: use Nacos config snapshot in ConfigData loader

DragonFSKY:

`NacosContextRefresher` stores the pushed Nacos config content in `NacosSnapshotConfigManager` before publishing the refresh event.

However, the ConfigData reload path uses `NacosConfigDataLoader#pullConfig`, which directly calls `ConfigService#getConfig(...)`. As a result, the pushed snapshot is not consumed in this path, and refresh may still read stale config from a Nacos cluster.

This PR makes `NacosConfigDataLoader` consume the pushed snapshot before falling back to `ConfigService#getConfig(...)`.

Addresses #4296

### Describe how you did it

`NacosConfigDataLoader#pullConfig` now first calls `NacosSnapshotConfigManager#getAndRemoveConfigSnapshot(dataId, group)`.

If the snapshot is `null` or empty, it keeps the previous behavior and calls `ConfigService#getConfig(...)`.

### Describe how to verify it

Added `NacosConfigDataLoaderTest` covering:

- snapshot exists: loader uses the snapshot and does not call remote `getConfig`
- snapshot is empty: loader falls back to remote `getConfig`

Test command:

```bash
./mvnw -pl spring-cloud-alibaba-starters/spring-alibaba-nacos-config -Dlicense.skip=true -Dcheckstyle.skip=true test
```

Result:

```text
Tests run: 22, Failures: 0, Errors: 0, Skipped: 0
```

### Special notes for reviews

The change is limited to the ConfigData loader path and preserves the existing fallback behavior when no valid snapshot is available.

### DIFF
--- a/spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/client/NacosPropertySourceBuilder.java
+++ b/spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/client/NacosPropertySourceBuilder.java
@@ -44,9 +44,17 @@ public class NacosPropertySourceBuilder {
 
 	private long timeout;
 
+	private @Nullable String namespace;
+
 	public NacosPropertySourceBuilder(ConfigService configService, long timeout) {
+		this(configService, timeout, null);
+	}
+
+	public NacosPropertySourceBuilder(ConfigService configService, long timeout,
+			@Nullable String namespace) {
 		this.configService = configService;
 		this.timeout = timeout;
+		this.namespace = namespace;
 	}
 
 	public long getTimeout() {
@@ -84,8 +92,9 @@ public class NacosPropertySourceBuilder {
 			String fileExtension) {
 		String data = null;
 		try {
-			String configSnapshot = NacosSnapshotConfigManager.getAndRemoveConfigSnapshot(dataId, group);
-			if (configSnapshot == null || configSnapshot.isEmpty()) {
+			String configSnapshot = NacosSnapshotConfigManager
+				.getAndRemoveConfigSnapshot(namespace, dataId, group);
+			if (configSnapshot == null) {
 				log.debug("get config from nacos, dataId: {}, group: {}", dataId, group);
 				data = configService.getConfig(dataId, group, timeout);
 			}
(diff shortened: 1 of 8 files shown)


xuxiaowei-com-cn:
- Thank you for your contribution and sign the agreement above.

DragonFSKY:
Thanks for the reminder. I have signed the CLA, and the CLA check is now passing.

uuuyuqi:
There are a few data consistency details that may be worth double-checking:

1. `NacosSnapshotConfigManager#getAndRemoveConfigSnapshot` currently does `get` first and then `remove`. With concurrent updates for the same `dataId + group`, an older reader could theoretically remove a newer snapshot that was just written by another thread. It may be safer to use a single `CONFIG_INFO_SNAPSHOT_MAP.remove(snapshotKey)` operation so reading and removal happen atomically.

2. The snapshot key currently only uses `dataId + "@" + group`, without namespace. If the same JVM has multiple Nacos namespaces with the same `dataId` and `group`, the snapshots may affect each other. It would be good to confirm whether this scenario needs to be supported; if yes, namespace should probably be part of the key.

3. An empty snapshot is currently treated as invalid and falls back to `ConfigService#getConfig(...)`. If updating a config to empty content is a valid use case, this can still read an old non-empty value from a Nacos cluster during server-side synchronization delay. It may be better to distinguish between "no snapshot exists" and "a snapshot exists with empty content" instead of checking only `config == null || config.isEmpty()`.

4. Snapshots do not have an expiration time. If a pushed snapshot is written but not consumed for some reason, a later load with the same key may read an older snapshot. The max-count limit helps with memory usage, but it does not prevent time-stale snapshot data. It may be worth adding timestamp metadata or at least evaluating the effect of unconsumed snapshots on later loads.

DragonFSKY:
Thank you for the detailed review and for pointing out these consistency cases. I pushed an update for the first three items:

1. `getAndRemoveConfigSnapshot(...)` now consumes the snapshot with a single `remove(...)` operation.
2. Snapshot keys are now namespace-aware, while keeping the existing `public` namespace behavior aligned with the default namespace.
3. Empty snapshots are now distinguished from missing snapshots, so an empty pushed config does not fall back to `ConfigService#getConfig(...)`.

I also added focused tests for the atomic consume behavior, namespace isolation/default namespace behavior, and empty-snapshot handling in both the ConfigData loader path and the property source builder path.

For point 4, I looked into it a bit more. I agree that an unconsumed snapshot can become time-stale and may be consumed by a later load with the same key. My current understanding is that this PR is mainly applying the existing snapshot mechanism from #3644 to the ConfigData loader path, which is the gap described in #4296.

I may be missing some context here, but adding timestamp metadata or a TTL seems to change the lifecycle semantics of `NacosSnapshotConfigManager`, and the expiration policy is not obvious to me because too short a TTL could make a slow refresh fall back to `ConfigService#getConfig(...)` again.

Would you prefer to include timestamp/expiration handling in this PR, or keep this PR focused and handle stale unconsumed snapshots as a separate follow-up? I would appreciate your guidance on the preferred direction.

Tests run locally:

```bash
./mvnw -pl spring-cloud-alibaba-starters/spring-alibaba-nacos-config -Dlicense.skip=true -Dcheckstyle.skip=true test
```

uuuyuqi (approved):
LGTM

PR #4321, opened 2026-05-08
feat: implement lightweight discovery probe for health check

elnafateh:

Fix Nacos discovery health checks to support lightweight probe()

- implement `DiscoveryClient#probe()` in `NacosDiscoveryClient`
- add lightweight `NacosServiceDiscovery.probe()` based on `NamingService.getServerStatus()`

This avoids forcing actuator discovery health checks through `getServices()` when users switch Spring Cloud Commons to probe mode.

#4294 

### Describe how to verify it
- add unit tests for healthy and failing probe cases

### Special notes for reviews
- document `spring.cloud.discovery.client.health-indicator.use-services-query=false` for Nacos discovery health checks

### DIFF
--- a/spring-cloud-alibaba-examples/nacos-example/readme-zh.md
+++ b/spring-cloud-alibaba-examples/nacos-example/readme-zh.md
@@ -328,6 +328,8 @@ SecretKey|spring.cloud.nacos.config.secret-key||
 是否开启 Nacos Config 的 Health Indicator|spring.nacos.config.health-indicator.enabled|false|
 是否开启 Nacos Discovery 的 Health Indicator|spring.cloud.nacos.discovery.health-indicator.enabled|false|
 
+对于服务发现健康检查，建议配置 `spring.cloud.discovery.client.health-indicator.use-services-query=false`，走更轻量的 `probe()` 路径。
+
 
 ### Spring Cloud Alibaba Nacos Discovery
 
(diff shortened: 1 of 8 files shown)


uuuyuqi (changes requested):
Thanks for this PR! The idea is solid and the implementation for the blocking path works correctly. I verified it locally with Arthas tracing — use-services-query=false properly routes health checks through probe() → NamingService.getServerStatus() instead of getServices().A couple of things to address before merge:

1. Missing NacosReactiveDiscoveryClient override
The PR only overrides probe() in NacosDiscoveryClient (blocking), but NacosReactiveDiscoveryClient also needs probe() and reactiveProbe() overridden. The default implementations in ReactiveDiscoveryClient still fall back to getServices(), which means WebFlux applications (e.g., Spring Cloud Gateway with Nacos discovery) won't benefit from this optimization when use-services-query=false.Since NacosReactiveDiscoveryClient already holds the same serviceDiscovery reference, the fix is straightforward — just delegate to serviceDiscovery.probe() the same way.

2. testProbe() should verify delegation
In NacosDiscoveryClientTests.testProbe(), since serviceDiscovery is a mock and void methods are no-ops by default, this test would still pass even if probe() were an empty method. Consider adding verify(serviceDiscovery).probe() to confirm the delegation actually happens.

elnafateh:
@uuuyuqi  The requested adjustments has been addressed.

- Added `probe()` and `reactiveProbe()` overrides to `NacosReactiveDiscoveryClient` so WebFlux apps also avoid falling back to `getServices()` when `use-services-query=false`
- Updated `NacosDiscoveryClientTests.testProbe()` to verify delegation
- Added reactive probe delegation and failure tests

Verified with targeted tests: 24 tests passed

uuuyuqi (approved):
LGTM

PR #4335, opened 2026-05-25
fix: resolve config refresh failure with spring.config.import on 2025.0.x

### LINKED ISSUE #4331: Nacos config refresh does not work with spring.config.import on 2025.0.x: PropertySource name mismatch and file extension lost in NacosPropertySourceRefreshListener
**Which Component**

Nacos Config (`spring-alibaba-nacos-config`)

---

**Describe the bug**

When using `spring-alibaba-nacos-config:2025.0.0.0` with Spring Boot `3.5.9` and `spring.config.import=nacos:...`, remote configuration can be loaded successfully at startup, but runtime refresh does not work correctly after the config is changed in Nacos.

After debugging the refresh flow, I found two related problems:

1. `NacosPropertySourceRefreshListener` cannot always locate the existing Nacos `PropertySource` in `Environment`, because the lookup key used in refresh does not match the actual property source name used by the ConfigData import path.
2. `NacosPropertySourceRefreshListener` rebuilds the property source with `"properties"` as the file extension, which is incorrect for configs imported as `yml/yaml/json/xml`.

As a result, the runtime `Environment` still exposes old values after a Nacos config change.

---

**Simplest demo**

I do not have a public demo repo at the moment, but the issue can be reproduced with a minimal Spring Boot application using:

- Spring Boot `3.5.9`
- `spring-alibaba-nacos-config:2025.0.0.0`
- `nacos-client:2.3.2`
- `spring.config.import=nacos:...yml?group=DEFAULT_GROUP&refreshEnabled=true`

Minimal configuration:

```yaml
spring:
  application:
    name: demo-app
  config:
    import:
      - "nacos:${spring.application.name}.yml?group=DEFAULT_GROUP&refreshEnabled=true"
  nacos:
    config:
      group: DEFAULT_GROUP
      refresh-enabled: true

### PR DESCRIPTION
**Pull request by wushiyuanmaimob**
  
  Fix two defects in `NacosPropertySourceRefreshListener` that prevented dynamic config refresh when using `spring.config.import=nacos:` on the 2025.0.x branch.
  
  **Defect 1: PropertySource name mismatch**
  - ConfigData path uses `group@dataId` naming (`NacosConfigDataLoader.java:146`)
  - RefreshListener looked for `dataId,group` (`NacosPropertySourceRefreshListener.java:103`)
  - Result: `target.get(sourceName)` always returned null, so refresh never happened
  
  **Defect 2: File extension hardcoded**
  - RefreshListener hardcoded `"properties"` as file extension (`NacosPropertySourceRefreshListener.java:108`)
  - Actual extension (yml/json/xml) was stored in `NacosItemConfig.suffix` but not propagated
  - Result: yml/json/xml configs were parsed as properties after refresh, causing errors
  
  Fixes #4331
  
  ### Describe how you did it
  
  1. Add `suffix` field to `NacosPropertySource` to preserve file extension
  2. Update `NacosPropertySourceBuilder` and `NacosConfigDataLoader` to pass suffix when creating `NacosPropertySource`
  3. Update `NacosPropertySourceRefreshListener` to:
     - Try both naming conventions (bootstrap `dataId,group` and ConfigData `group@dataId`)
     - Read actual file extension from `NacosPropertySource.getSuffix()` instead of hardcoding `"properties"`
     
  ### Describe how to verify it
  
  1. Set up a Spring Boot 3.x app with `spring.config.import=nacos:test-config.yml`
  2. Change the config value in Nacos
  3. Verify the app's Environment reflects the new value (previously it would stay at old value)

  Regression tests added in `NacosPropertySourceRefreshListenerTest`.                                                                               

  ### Special notes for reviews

  - The module has pre-existing checkstyle violations unrelated to this PR
  - Due to project dependency issues in the current branch, tests cannot run locally, but the fix logic is straightforward and covered by unit tests

### DIFF
--- a/spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/refresh/NacosPropertySourceRefreshListener.java
+++ b/spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/refresh/NacosPropertySourceRefreshListener.java
@@ -100,12 +100,32 @@ public class NacosPropertySourceRefreshListener implements BeanPostProcessor, Sm
 
 				NacosPropertySourceBuilder nacosPropertySourceBuilder = new NacosPropertySourceBuilder(nacosConfigManager.getConfigService(), nacosConfigManager.getNacosConfigProperties()
 						.getTimeout());
-				String sourceName = String.join(NacosConfigProperties.COMMAS, event.dataId, event.group);
+
+				// Try both naming conventions: bootstrap path (dataId,group) and ConfigData path (group@dataId)
+				String bootstrapName = String.join(NacosConfigProperties.COMMAS, event.dataId, event.group);
+				String configDataName = event.group + "@" + event.dataId;
+
 				ConfigurableEnvironment environment = ((ConfigurableApplicationContext) applicationContext).getEnvironment();
 				MutablePropertySources target = environment.getPropertySources();
-				PropertySource<?> prevpropertySource = target.get(sourceName);
+
+				PropertySource<?> prevpropertySource = target.get(bootstrapName);
+				String sourceName = bootstrapName;
+
+				// If not found with bootstrap naming, try ConfigData naming
+				if (prevpropertySource == null) {
+					prevpropertySource = target.get(configDataName);
+					sourceName = configDataName;
+				}
+
 				if (prevpropertySource instanceof NacosPropertySource) {
-					NacosPropertySource newProperSource = nacosPropertySourceBuilder.build(event.getDataId(), event.getGroup(), "properties", ((NacosPropertySource) prevpropertySource).isRefreshable());
+					NacosPropertySource prevNacosSource = (NacosPropertySource) prevpropertySource;
+					// Use the actual suffix from the previous source instead of hardcoding "properties"
+					String fileExtension = prevNacosSource.getSuffix();
+					if (fileExtension == null || fileExtension.isEmpty()) {
+						fileExtension = "properties";
+					}
+					NacosPropertySource newProperSource = nacosPropertySourceBuilder.build(
+							event.getDataId(), event.getGroup(), fileExtension, prevNacosSource.isRefreshable());
 					target.replace(sourceName, newProperSource);
 					log.info("Replace Nacos Property Source : " + sourceName);
 
--- a/spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/client/NacosPropertySource.java
+++ b/spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/client/NacosPropertySource.java
@@ -55,19 +55,31 @@ public class NacosPropertySource extends MapPropertySource {
 	 */
 	private final boolean isRefreshable;
 
+	/**
+	 * File extension (suffix) of the config data (e.g., "properties", "yml", "json").
+	 */
+	private final String suffix;
+
 	NacosPropertySource(String group, String dataId, Map<String, Object> source,
-			Date timestamp, boolean isRefreshable) {
+			Date timestamp, boolean isRefreshable, String suffix) {
 		super(String.join(NacosConfigProperties.COMMAS, dataId, group), source);
 		this.group = group;
 		this.dataId = dataId;
 		this.timestamp = timestamp;
 		this.isRefreshable = isRefreshable;
+		this.suffix = suffix;
 	}
 
 	public NacosPropertySource(List<PropertySource<?>> propertySources, String group,
 			String dataId, Date timestamp, boolean isRefreshable) {
 		this(group, dataId, getSourceMap(group, dataId, propertySources), timestamp,
-				isRefreshable);
+				isRefreshable, "properties");
+	}
+
+	public NacosPropertySource(List<PropertySource<?>> propertySources, String group,
+			String dataId, Date timestamp, boolean isRefreshable, String suffix) {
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by bengbengbalabalabeng** [comment]
Will the current PR see any further progress?

File: spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/refresh/NacosPropertySourceRefreshListener.java [hunk 1]
```diff
@@ -100,12 +100,32 @@ public void handle(NacosConfigRefreshEvent event) {
 
 				NacosPropertySourceBuilder nacosPropertySourceBuilder = new NacosPropertySourceBuilder(nacosConfigManager.getConfigService(), nacosConfigManager.getNacosConfigProperties()
 						.getTimeout());
-				String sourceName = String.join(NacosConfigProperties.COMMAS, event.dataId, event.group);
+
```
bengbengbalabalabeng, line 103: Based on the fix in #4341, it looks like the naming mismatch issue should no longer occur.

oss-sentinel-ai (commented):
﻿### Summary
This PR fixes two real defects in the config refresh path when using `spring.config.import=nacos:`: (1) PropertySource name mismatch between ConfigData (`group@dataId`) and bootstrap (`dataId,group`) naming conventions, and (2) hardcoded `"properties"` suffix during refresh. The dual-name lookup approach is pragmatic. However, there are a few concerns worth addressing before merge.

### Findings
- **[Critical]** `NacosPropertySourceRefreshListener.java:125` — When the ConfigData path is taken, `sourceName` is set to `configDataName` (`group@dataId`), but the rebuilt `NacosPropertySource` from `nacosPropertySourceBuilder.build()` constructs its internal name as `dataId,group` (bootstrap convention). After `target.replace(sourceName, newProperSource)`, the map key is `group@dataId` but the object's `getName()` returns `dataId,group`. This inconsistency can cause subtle bugs in downstream code that relies on `NacosPropertySource.getName()` matching its key in the `PropertySources` map. Consider passing the original `sourceName` into the builder or using a constructor that accepts an explicit name.
- **[Warning]** This PR and #4349 both fix the same hardcoded-suffix bug with different approaches (immutable constructor param vs. mutable setter). They will conflict if both are merged. Please coordinate with the other PR author to align on a single approach.
- **[Info]** Test file path `src/test/java/com.alibaba.cloud.nacos/refresh/` uses dots instead of slashes for the package directory. This may cause test discovery issues depending on the build configuration.

### Suggestions
```java
// In NacosPropertySourceRefreshListener, when rebuilding for ConfigData path:
// Ensure the rebuilt source name matches the map key
String sourceName = bootstrapName;
NacosPropertySource existing = (NacosPropertySource) target.get(sourceName);
if (existing == null) {
    sourceName = configDataName;
    existing = (NacosPropertySource) target.get(sourceName);
}
// existing.getSuffix() now carries the correct extension — good.
// But verify the rebuilt source name matches sourceName.
```

---

uuuyuqi:
Thanks for the contribution. I rechecked the current `2025.0.x` and `2025.1.x` branches. The two problems addressed by this PR now have different statuses:

1. The ConfigData-loaded `PropertySource` could not be replaced correctly by the refresh listener. This part has already been backported to `2025.0.x` through #4346. Therefore, the dual lookup for `dataId,group` and `group@dataId` in this PR is no longer needed.
2. The original suffix is lost when refreshing non-properties configurations. This has been fully fixed in `2025.1.x` through #4349, but it has not yet been backported to `2025.0.x`.

Please rebase this PR onto the latest `2025.0.x` and revise it into a backport of #4349. The backport should preserve the original suffix on each replacement `NacosPropertySource` and retain the regression test that verifies two consecutive refreshes continue to use the same suffix.

The #4349 commit cannot be cherry-picked into `2025.0.x` without adaptation. The main branch-specific differences are:

- `NacosPropertySourceBuilder` in `2025.0.x` does not have the `namespace` constructor argument.
- `2025.0.x` does not include the JSpecify `@Nullable` dependency.
- The test copyright headers need to follow the Checkstyle rules used by `2025.0.x`.
- The snapshot regression test carried over from #4319 is not applicable to the current `2025.0.x` branch and does not need to be included in this backport.

After these adjustments, please make sure the tests cover yml, json, the null-suffix fallback, and consecutive refreshes. We can continue the review once the PR has been updated.

uuuyuqi (changes requested):
see https://github.com/alibaba/spring-cloud-alibaba/pull/4335#issuecomment-5391932263

oss-sentinel-ai (commented):
Thanks @uuuyuqi for the detailed guidance. Updating my earlier review in light of it:

- **[Warning] overlap with #4349 — resolved:** the maintainers decided this PR should be revised into a backport of #4349, so no coordination between competing approaches is needed anymore.
- **[Critical] rebuilt source name mismatch — superseded:** since #4346 already landed the ConfigData replacement fix on `2025.0.x` (merged 2026-06-21), the dual `dataId,group` / `group@dataId` lookup in this PR is no longer needed once rebased.
- **[Info] test directory path — still applies:** `src/test/java/com.alibaba.cloud.nacos/refresh/` uses dots instead of slashes; if the test is carried over into the backport, please use `src/test/java/com/alibaba/cloud/nacos/refresh/` so test discovery works reliably.

Remaining work per the maintainer's request: rebase onto the latest `2025.0.x` and revise this PR into a backport of #4349 (merged on `2025.1.x` 2026-07-31) — preserve the original suffix on each replacement `NacosPropertySource`, retain the consecutive-refresh regression test, adapt for the missing `namespace` constructor argument and the absent JSpecify `@Nullable` dependency, follow the `2025.0.x` Checkstyle copyright headers, and skip the #4319 snapshot regression test. Test coverage should include yml, json, the null-suffix fallback, and consecutive refreshes.

Happy to re-review once the PR is updated.

---

Issue #4337, opened 2026-05-27
[Bug] NacosPropertySource replacement failed


bengbengbalabalabeng:
**Which Component**

```xml
<dependency>
	<groupId>com.alibaba.cloud</groupId>
	<artifactId>spring-alibaba-nacos-config</artifactId>
	<version>2025.1.0.0</version>
</dependency>
```

spring-alibaba-nacos-config + Spring Boot v4.0.6 (without any Cloud)

**Describe the bug**

In `NacosPropertySourceRefreshListener#handle(NacosConfigRefreshEvent)`, the `List<PropertySource<?>>` obtained via `environment.getPropertySources()` does not contain any `NacosPropertySource` instances.

Therefore, the following logic will never be triggered:

```java
if (prevpropertySource instanceof NacosPropertySource) {
    // ...
    target.replace(sourceName, newProperSource);
    // ...
}
```

After the configuration refresh event is triggered, the properties in the Environment are not replaced, and `Environment#getProperty()` always returns the old value.

> Note: We will not discuss issues related to #4331 here for now.

**To Reproduce**

1. Register the `NacosPropertySourceRefreshListener` Bean  
2. Load configuration using `spring.config.import=nacos:xxx`  
3. Modify the configuration in the Nacos Console  
4. Call `Environment#getProperty()` multiple times and find that the value does not update

**Root Cause**

`NacosConfigDataLoader#doLoad → pullConfig → NacosDataParserHandler#parseNacosData`

`NacosDataParserHandler` loads the following `PropertySourceLoader`s via SPI:

-  `NacosJsonPropertySourceLoader`
-  `NacosXmlPropertySourceLoader`
-  `PropertiesPropertySourceLoader`
-  `YamlPropertySourceLoader`

These loaders **all return `OriginTrackedMapPropertySource`**, not `NacosPropertySource`.

Therefore, the PropertySource ultimately added to the `Environment` is of type `OriginTrackedMapPropertySource` rather than `NacosPropertySource`. This causes the type check in `NacosPropertySourceRefreshListener` to always fail.

elnafateh:
this issue has been worked out by the above PR #4341

oss-sentinel-ai:
**Issue Evaluation**

Category: `kind/bug` | Status: **Confirmed**

The reported issue has been verified against the current codebase.

[...]

PR #4341, opened 2026-06-15
fix(nacos): NacosPropertySource refresh replacement

elnafateh:

Expose Nacos config data as NacosPropertySource so refresh events can replace the environment property source correctly. Adds regression assertions for the loaded property source type and name.

ISSUE #4337


uuuyuqi (approved):
LGTM. Non-properties refresh still needs a follow-up fix to preserve and reuse the original suffix/file extension during rebuild.

uuuyuqi:
Thanks for the fix. I checked the 2025.0.x branch and it appears to have the same issue, so we will need a similar PR for 2025.0.x as well.

The yml / yaml / json / xml refresh case is a separate follow-up issue. Contributions for that fix are welcome.

elnafateh:
@uuuyuqi My Pleasure.. So the follow ups PR will be

1. #4349
2.  #4346

elnafateh:
@uuuyuqi follow up PRs are now achieved

PR #4346, opened 2026-06-17
fix(nacos): backport NacosPropertySource refresh replacement to 2025.0.x

### LINKED ISSUE #4337: [Bug] NacosPropertySource replacement failed

elnafateh:
Issue #4337

### DIFF
--- a/spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/configdata/NacosConfigDataLoader.java
+++ b/spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/configdata/NacosConfigDataLoader.java
@@ -18,6 +18,7 @@ package com.alibaba.cloud.nacos.configdata;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Date;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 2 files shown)


xuxiaowei-com-cn (commented):
LGTM

xuxiaowei-com-cn (approved):
LGTM

PR #4349, opened 2026-06-21
fix(nacos): preserve file extension for yml/json/xml config refresh

### LINKED ISSUE #4337: [Bug] NacosPropertySource replacement failed

elnafateh:

When NacosPropertySourceRefreshListener handles a NacosConfigRefreshEvent,
it was calling NacosPropertySourceBuilder.build() with a hardcoded
'properties' suffix, causing yml/yaml/json/xml Nacos configs to be
reparsed as .properties format on refresh, producing empty or wrong values.

Fix: store the original file extension (suffix) on NacosPropertySource at
load time in NacosConfigDataLoader, then read it back in the refresh
listener to pass the correct suffix to NacosDataParserHandler.

Also extract nacosPropertySourceBuilder as a field on the listener
to allow lazy initialisation and testability.

Falls back to 'properties' if suffix is null for backwards compatibility.

Closes: #4337

Follow-up to #4341 (fix for `.properties` refresh).
Fixes #4337.


uuuyuqi (changes requested):
Thanks for the follow-up fix. The overall direction looks right: `yml / yaml / json / xml `are already supported by the existing PropertySourceLoader infrastructure, so this PR should mainly preserve the original suffix across the refresh replacement path.

I think there is still one issue before merging. The initial `ConfigData` load sets the suffix on `NacosPropertySource`, and the listener uses it correctly for the first refresh. However, the replacement `NacosPropertySource` created by `NacosPropertySourceBuilder.build(...)` does not set the suffix back. After the first `target.replace(...)`, the environment contains a new `NacosPropertySource` with `suffix == null`, so the next refresh falls back to `properties`.

elnafateh:
Fix: call setSuffix(fileExtension) on the built NacosPropertySource
before returning it, ensuring the suffix is preserved across all
subsequent refresh cycles

uuuyuqi:
Thanks for the update. The suffix propagation fix looks good to me. I have one remaining request and one additional test suggestion:

1. Removing the misplaced duplicate test file makes sense, but please preserve the existing `buildWhenSnapshotIsEmptyThenDoesNotUseRemoteConfig` test by moving it into the correctly located `NacosPropertySourceBuilderTest`. Otherwise, this PR unintentionally drops the regression coverage for the empty-snapshot behavior introduced in #4319.
2. Please also consider adding a regression test that performs two consecutive refreshes using a non-properties format, such as `yml` or `json`, and verifies that both refreshes use the original suffix. Ideally, the second refresh should operate on the replacement `NacosPropertySource` produced by the first refresh. This would directly cover the issue raised in my previous review and ensure that subsequent refreshes do not fall back to `properties`.

uuuyuqi (changes requested):
see https://github.com/alibaba/spring-cloud-alibaba/pull/4349#pullrequestreview-4544136001

oss-sentinel-ai (approved):
### Summary
This PR correctly fixes a real bug where Nacos config refresh was reparsing yml/json/xml configs as `.properties` format due to a hardcoded suffix. The approach of storing the original file extension on `NacosPropertySource` and reading it back during refresh is clean and backward-compatible. Good test coverage across all supported formats.

### Findings
- **[Warning]** `NacosPropertySourceRefreshListener.java:57` — Lazy-init of `nacosPropertySourceBuilder` without synchronization; consider local variable or `volatile`
- **[Info]** `NacosPropertySource.java:59` — Document the null→"properties" fallback in Javadoc

---

oss-sentinel-ai (approved):
### Summary
This PR correctly fixes the bug where Nacos config refresh was reparsing yml/json/xml configs as `.properties` format. The new commit addresses all previous feedback:

- **My previous findings**: Both resolved — `builder` is now a local variable (no thread safety concern), and the `suffix` field has proper Javadoc documenting the null→`properties` fallback.
- **Maintainer requests**: Both addressed — `buildWhenSnapshotIsEmptyThenDoesNotUseRemoteConfig` test preserved in correctly located `NacosPropertySourceBuilderTest`, and `consecutiveRefreshesPreserveYmlSuffix` test added to verify suffix propagation across multiple refresh cycles.

Test coverage is comprehensive: unit tests for `NacosPropertySource` suffix field, builder suffix preservation across all formats, and refresh listener behavior including the consecutive refresh scenario.

LGTM — ready for maintainer review.

---

uuuyuqi (approved):
LGTM

PR #4350, opened 2026-06-21
fix(nacos): skip loopback addresses for network interface

### LINKED ISSUE #4342: NacosDiscoveryProperties picks loopback IP when network-interface is configured
**Which Component**
Nacos Discovery (spring-cloud-starter-alibaba-nacos-discovery)

**Describe the bug**

In NacosDiscoveryProperties.init(), the condition that filters addresses on a given network interface has an operator-precedence bug: && binds tighter than ||, so the loopback check only applies to IPv6 addresses. As a result, when a user configures spring.cloud.nacos.discovery.network-interface, an IPv4 loopback address (e.g. 127.0.0.1) may be selected as the registration IP, making the service unreachable from other nodes in the cluster.
The current code:
`// File: NacosDiscoveryProperties.java (line 309-314)
if (currentAddress instanceof Inet4Address
        || currentAddress instanceof Inet6Address
        && !currentAddress.isLoopbackAddress()) {
    ip = currentAddress.getHostAddress();
    break;
}`

is parsed by Java as:

`if (currentAddress instanceof Inet4Address
        || (currentAddress instanceof Inet6Address && !currentAddress.isLoopbackAddress()))`

中文原文：由于 && 优先级高于 ||，IPv4 的 loopback 地址不会被排除，只有 IPv6 的 loopback 地址会被排除。当用户指定网络接口时，可能选到 127.0.0.1 作为注册 IP。

**Simplest demo**

Run any nacos-discovery example (e.g. nacos-discovery-provider-example) on a machine where the chosen network interface contains both a loopback IPv4 entry and a regular IPv4 entry, and set:

spring:
  cloud:
    nacos:
      discovery:
        network-interface: lo0   # or any interface where loopback is enumerated first

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**
OS: any (verified by static analysis)
JDK: 17
Spring Cloud Alibaba: 2025.1.0.1-SNAPSHOT

hutiefang76:
Fixes #4342

## Summary

This fixes the address selection when `spring.cloud.nacos.discovery.network-interface` is configured. The previous condition was parsed as `IPv4 || (IPv6 && !loopback)`, so an IPv4 loopback address could be selected before a real address on the same interface.

Changes included:

- apply the loopback check to both IPv4 and IPv6 candidates in `NacosDiscoveryProperties.init()`
- add a regression test where the mocked network interface returns `127.0.0.1` before `192.168.1.10`
- align the same condition in the related network-interface test helper

## Verification

- Red check before the production fix: `NacosDiscoveryPropertiesTests#initShouldSkipLoopbackAddressFromNetworkInterface` failed because the selected IP was `127.0.0.1` instead of `192.168.1.10`
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) ./mvnw -pl spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery -am -Dtest=NacosDiscoveryPropertiesTests#initShouldSkipLoopbackAddressFromNetworkInterface -Dsurefire.failIfNoSpecifiedTests=false test`
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) ./mvnw -pl spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery -am -Dtest=NacosDiscoveryPropertiesTests,NacosAutoServiceRegistrationIpNetworkInterfaceTests -Dsurefire.failIfNoSpecifiedTests=false test`
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) ./mvnw -pl spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery -am test`
- `git diff --check`

The current `2025.1.x` branch requires JDK 21 for this module build path, so the commands above were run with JDK 21.

### DIFF
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/main/java/com/alibaba/cloud/nacos/NacosDiscoveryProperties.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/main/java/com/alibaba/cloud/nacos/NacosDiscoveryProperties.java
@@ -305,8 +305,8 @@ public class NacosDiscoveryProperties {
 				Enumeration<InetAddress> inetAddress = netInterface.getInetAddresses();
 				while (inetAddress.hasMoreElements()) {
 					InetAddress currentAddress = inetAddress.nextElement();
-					if (currentAddress instanceof Inet4Address
-							|| currentAddress instanceof Inet6Address
+					if ((currentAddress instanceof Inet4Address
+							|| currentAddress instanceof Inet6Address)
 							&& !currentAddress.isLoopbackAddress()) {
 						ip = currentAddress.getHostAddress();
 						break;
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/test/java/com/alibaba/cloud/nacos/NacosDiscoveryPropertiesTests.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/test/java/com/alibaba/cloud/nacos/NacosDiscoveryPropertiesTests.java
@@ -16,6 +16,10 @@
 
 package com.alibaba.cloud.nacos;
 
+import java.net.InetAddress;
+import java.net.NetworkInterface;
+import java.util.Collections;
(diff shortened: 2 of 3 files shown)


uuuyuqi (approved):
LGTM

PR #4351, opened 2026-06-22
fix(sentinel): map fallback methods to implementations

### LINKED ISSUE #4345: SentinelInvocationHandler.toFallbackMethod maps interface methods to themselves
**Which Component**
Sentinel

**Describe the bug**
toFallbackMethod is supposed to associate each Feign interface method with the corresponding method on the fallback class, so that the correct method (potentially with overrides, AOP advice, or extra annotations) is invoked when fallback is triggered. The current implementation maps each interface method to itself:

```
// File: SentinelInvocationHandler.java (lines 180-187)
static Map<Method, Method> toFallbackMethod(Map<Method, MethodHandler> dispatch) {
    Map<Method, Method> result = new LinkedHashMap<>();
    for (Method method : dispatch.keySet()) {
        method.setAccessible(true);
        result.put(method, method);   // <-- mapped to itself
    }
    return result;
}
```

While this happens to work for plain reflective dispatch on interface methods, it bypasses any method-level customization on the fallback class and makes the setAccessible(true) call meaningless for public interface methods.
中文原文：fallback 方法被映射到接口方法本身而非 fallback 类上对应的方法，绕过了 fallback 实现类上的覆盖、AOP 增强等逻辑。

**Simplest demo**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

hutiefang76:

This PR fixes the Sentinel Feign fallback method mapping so each Feign interface method can resolve to the corresponding method declared on the fallback implementation class.

Previously, `SentinelInvocationHandler.toFallbackMethod` mapped each interface method back to itself. Invocation still worked for simple fallback objects, but the mapping did not preserve the actual fallback implementation method.

Fixes #4345

### Describe how you did it

- Added a fallback-type-aware overload of `toFallbackMethod`.
- Passed the actual fallback instance class from `SentinelFeign` when a direct fallback is configured.
- Resolved the fallback method against the actual fallback object before invocation, which also keeps `fallbackFactory` results aligned with the implementation class.
- Added a focused unit test to verify that an interface method maps to the fallback implementation method.

### Describe how to verify it

```bash
./mvnw -pl spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-sentinel -am -Dtest=SentinelInvocationHandlerTests -Dsurefire.failIfNoSpecifiedTests=false test
./mvnw -pl spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-sentinel -am test -Dsurefire.failIfNoSpecifiedTests=false
git diff --check
```

### Special notes for reviews

The fallback invocation path still creates the fallback object once per fallback handling, as before. The added resolution step only changes which `Method` object is invoked.

### DIFF
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-sentinel/src/main/java/com/alibaba/cloud/sentinel/feign/SentinelFeign.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-sentinel/src/main/java/com/alibaba/cloud/sentinel/feign/SentinelFeign.java
@@ -127,7 +127,8 @@ public final class SentinelFeign {
 						fallbackInstance = getFromContext(beanName, "fallback", fallback,
 								target.type());
 						return new SentinelInvocationHandler(target, dispatch,
-								new FallbackFactory.Default(fallbackInstance));
+								new FallbackFactory.Default(fallbackInstance),
+								fallbackInstance.getClass());
 					}
 					if (void.class != fallbackFactory) {
 						fallbackFactoryInstance = (FallbackFactory) getFromContext(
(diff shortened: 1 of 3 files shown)


hutiefang76:
Closing this after re-checking the behavior evidence. A local probe shows that invoking the interface Method on a Spring AOP fallback proxy still triggers the fallback advice, so this change does not currently prove a real user-visible bug beyond changing the stored Method object. I prefer to withdraw it rather than keep an unclear fix open.

PR #4360, opened 2026-07-14
fix: resolve sentinel feign fallback methods

### LINKED ISSUE #4345: SentinelInvocationHandler.toFallbackMethod maps interface methods to themselves

jay666mnj:

Fix Sentinel Feign fallback method resolution so fallback invocations use the concrete fallback implementation method instead of the Feign interface method.

Fixes #4345

### Describe how you did it

Pass the fallback implementation class to `SentinelInvocationHandler` when a direct fallback instance is configured.

Resolve fallback methods by name and parameter types against the fallback implementation class, and resolve/cache the actual fallback instance method for `FallbackFactory` cases.

Added a unit test to verify fallback class method mapping.

### Describe how to verify it

- `git diff --check`

I also tried to run the targeted Maven test, but local verification was blocked before test execution because this machine uses JDK 17.0.10 while the project requires JDK 21.0.8+.

### Special notes for reviews

NONE

### DIFF
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-sentinel/src/main/java/com/alibaba/cloud/sentinel/feign/SentinelInvocationHandler.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-sentinel/src/main/java/com/alibaba/cloud/sentinel/feign/SentinelInvocationHandler.java
@@ -126,7 +126,9 @@ public class SentinelInvocationHandler implements InvocationHandler {
 							if (fallbackMethod == null) {
 								throw new IllegalStateException("Fallback method not found for method: " + method);
 							}
-							Object fallbackResult = fallbackMethod.invoke(fallbackFactory.create(ex), args);
+							Object fallback = fallbackFactory.create(ex);
+							validateFallbackInstance(fallback);
+							Object fallbackResult = fallbackMethod.invoke(fallback, args);
 							return fallbackResult;
 						}
 						catch (IllegalAccessException e) {
@@ -186,4 +188,12 @@ public class SentinelInvocationHandler implements InvocationHandler {
 		return result;
 	}
 
+	private void validateFallbackInstance(Object fallback) {
+		if (!target.type().isInstance(fallback)) {
+			throw new IllegalStateException(String.format(
+					"Incompatible fallback instance. FallbackFactory returned %s, which is not assignable to %s for feign client %s",
+					fallback == null ? null : fallback.getClass(), target.type(),
+					target.name()));
+		}
+	}
 }
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-sentinel/src/test/java/com/alibaba/cloud/sentinel/SentinelFeignFallbackFactoryTests.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-sentinel/src/test/java/com/alibaba/cloud/sentinel/SentinelFeignFallbackFactoryTests.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2013-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.alibaba.cloud.sentinel;
+
+import com.alibaba.cloud.sentinel.feign.SentinelFeignAutoConfiguration;
+import com.alibaba.csp.sentinel.slots.block.RuleConstant;
+import com.alibaba.csp.sentinel.slots.block.flow.FlowRule;
+import com.alibaba.csp.sentinel.slots.block.flow.FlowRuleManager;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.cloud.openfeign.EnableFeignClients;
+import org.springframework.cloud.openfeign.FallbackFactory;
+import org.springframework.cloud.openfeign.FeignClient;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+import org.springframework.web.bind.annotation.RequestMapping;
+
+import static java.util.Collections.singletonList;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+@SpringJUnitConfig
+@SpringBootTest(classes = SentinelFeignFallbackFactoryTests.TestConfig.class,
+		properties = "feign.sentinel.enabled=true")
+class SentinelFeignFallbackFactoryTests {
+
+	@Autowired
+	private InvalidFallbackService invalidFallbackService;
+
+	@BeforeEach
+	void setUp() {
+		FlowRule rule = new FlowRule();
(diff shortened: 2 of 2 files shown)


xuxiaowei-com-cn:
Thank you for your first contribution. Please sign the CLA agreement above.

jay666mnj:
my pleasure! I have signed it~


         原始邮件
         
       
发件人：徐晓伟 ***@***.***&gt;
发件时间：2026年7月14日 22:55
收件人：alibaba/spring-cloud-alibaba ***@***.***&gt;
抄送：yijie zhao ***@***.***&gt;, Author ***@***.***&gt;
主题：Re: [alibaba/spring-cloud-alibaba] fix: resolve sentinel feign fallback methods (PR #4360)



xuxiaowei-com-cn left a comment (alibaba/spring-cloud-alibaba#4360)

Thank you for your first contribution. Please sign the CLA agreement above.

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS&nbsp;and Android. Download it today!

uuuyuqi:
Thanks for the contribution. After reviewing #4345 and this change, I don’t think we should change the fallback method resolution without a concrete reproducer.

A configured fallback is required to implement the Feign client interface. For instance methods, `Method.invoke` performs dynamic method lookup, so invoking the interface `Method` on a valid fallback object already dispatches to the implementation override. It also enters the proxy chain when the fallback object is proxied.

Spring Cloud OpenFeign’s own [`FeignCircuitBreakerInvocationHandler`](https://github.com/spring-cloud/spring-cloud-openfeign/blob/main/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignCircuitBreakerInvocationHandler.java) intentionally keeps the `dispatch method -> same interface method` mapping. Its comment explains that this stores an accessible method copy for package-private interfaces and avoids repeated reflection.

I also have several concerns about the current implementation:

- For `FallbackFactory`, resolving a method only by name and parameter types weakens the Feign type contract. An object that does not implement the Feign interface but happens to expose a matching method could be invoked instead of failing fast.
- `fallback.getClass()` may return a JDK or CGLIB proxy class, so it does not reliably identify the user implementation method or its annotations.
- The method is resolved on every fallback invocation and then written into a shared `LinkedHashMap`. The cache therefore does not avoid reflection and introduces unnecessary shared mutation.
- The added test only verifies `Method.getDeclaringClass()`. It does not demonstrate a fallback, AOP, or annotation scenario that fails before this change and passes afterward.

Could you provide a minimal regression test that fails on the base branch and passes with this patch, ideally reproducing the claimed AOP or annotation issue? Without such a reproducer, I would prefer to keep the existing interface-method invocation.

If the goal is to provide clearer validation for an invalid object returned by `FallbackFactory`, an explicit `target.type().isInstance(fallback)` check would preserve the Feign interface contract and fail with a clear error.

jay666mnj:
Thanks for the detailed review. I have updated the PR based on your suggestion.


The fallback method re-resolution logic has been removed, so the existing OpenFeign interface-method invocation behavior is preserved.


This PR now only adds an explicit validation for the object returned by FallbackFactory#create. If the returned fallback is not an instance of the Feign client interface, SentinelInvocationHandler fails fast with a clear IllegalStateException.


I also replaced the previous test with a JUnit 5 regression test:
SentinelFeignFallbackFactoryTests#invalidFallbackFactoryReturnTypeShouldFailFast


Verified locally with JDK 21:
mvn -pl spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-sentinel -am -Denforcer.skip=true -DskipCheckstyle=true -Dcheckstyle.skip=true -Dmaven.compiler.proc=none -Dtest=SentinelFeignFallbackFactoryTests#invalidFallbackFactoryReturnTypeShouldFailFast -Dsurefire.failIfNoSpecifiedTests=false test


Result:
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS


         原始邮件
         
       
发件人：uuuyuqi ***@***.***&gt;
发件时间：2026年7月15日 14:25
收件人：alibaba/spring-cloud-alibaba ***@***.***&gt;
抄送：yijie zhao ***@***.***&gt;, Author ***@***.***&gt;
主题：Re: [alibaba/spring-cloud-alibaba] fix: resolve sentinel feign fallback methods (PR #4360)



uuuyuqi left a comment (alibaba/spring-cloud-alibaba#4360)

Thanks for the contribution. After reviewing #4345&nbsp;and this change, I don’t think we should change the fallback method resolution without a concrete reproducer.

A configured fallback is required to implement the Feign client interface. For instance methods, Method.invoke&nbsp;performs dynamic method lookup, so invoking the interface Method&nbsp;on a valid fallback object already dispatches to the implementation override. It also enters the proxy chain when the fallback object is proxied.

Spring Cloud OpenFeign’s own FeignCircuitBreakerInvocationHandler&nbsp;intentionally keeps the dispatch method -&gt; same interface method&nbsp;mapping. Its comment explains that this stores an accessible method copy for package-private interfaces and avoids repeated reflection.

I also have several concerns about the current implementation:

For FallbackFactory, resolving a method only by name and parameter types weakens the Feign type contract. An object that does not implement the Feign interface but happens to expose a matching method could be invoked instead of failing fast.

fallback.getClass()&nbsp;may return a JDK or CGLIB proxy class, so it does not reliably identify the user implementation method or its annotations.

The method is resolved on every fallback invocation and then written into a shared LinkedHashMap. The cache therefore does not avoid reflection and introduces unnecessary shared mutation.

The added test only verifies Method.getDeclaringClass(). It does not demonstrate a fallback, AOP, or annotation scenario that fails before this change and passes afterward.

Could you provide a minimal regression test that fails on the base branch and passes with this patch, ideally reproducing the claimed AOP or annotation issue? Without such a reproducer, I would prefer to keep the existing interface-method invocation.

If the goal is to provide clearer validation for an invalid object returned by FallbackFactory, an explicit target.type().isInstance(fallback)&nbsp;check would preserve the Feign interface contract and fail with a clear error.

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS&nbsp;and Android. Download it today!

uuuyuqi:
Thank you for responding quickly to the review feedback, adding a regression test, and taking the time to verify the change locally. The latest version removes the fallback method re-resolution logic and preserves OpenFeign’s existing interface-method invocation behavior, which is the right direction.

However, after further evaluation, we still do not plan to merge this PR, mainly for the following reasons:

1. The current change no longer addresses the fallback method resolution problem described in #4345, so the PR title, description, and `Fixes #4345` reference no longer match the actual code changes.
2. The new validation targets a `FallbackFactory` that bypasses generic type safety by using a raw type and returns an object that does not implement the Feign client interface. With the normal, supported `FallbackFactory<T>` usage, the return type is already constrained by generics. For such an invalid object, the existing implementation also throws during reflective invocation rather than proceeding silently.
3. Therefore, the practical benefit of the current change is primarily a clearer exception type and error message for a non-standard configuration. At present, we do not have a real user report or a reproducible supported-use case demonstrating the need for this additional validation. That benefit is not sufficient to justify the additional implementation and test maintenance.

For these reasons, we do not plan to merge this PR and suggest closing it. If a real user report later demonstrates the need for clearer type validation here, we can revisit it as a separate issue with a concrete reproduction.

Thank you again for your enthusiasm, the time you invested, and your thoughtful response to the review feedback. You are very welcome to continue contributing to other Spring Cloud Alibaba issues. Issues with clear reproduction steps and expected behavior are especially valuable; if the intended solution is uncertain, please also feel free to discuss the approach in the issue before implementation. We look forward to your future contributions.

PR #4361, opened 2026-07-15
fix(rocketmq): log warning when FLAG or DELAY header is not numeric (2025.0.x backport of #4330)

daguimu:

Backport of #4330 to `2025.0.x`, as requested by @uuuyuqi in https://github.com/alibaba/spring-cloud-alibaba/pull/4330#issuecomment-4976150069.

`RocketMQMessageConverterSupport#getAndWrapMessage` on this branch still swallows any `Integer.parseInt` failure on the `MQ_FLAG` and `DELAY_TIME_LEVEL` Spring Messaging headers with `catch (Exception ignored) {}`, silently falling back to `flag=0` and `delayLevel=0`. The most user-visible consequence is on the delayed-delivery path: a misconfigured `DELAY_TIME_LEVEL` (e.g. `"abc"` from a SpEL expression that returned a String) turns a delayed message into an immediate send with no exception, no log line, and no signal to the caller.

NONE (backport of #4330)

### Describe how you did it

Same change as #4330, adapted to this branch:

- Extracted the parsing into a `parseHeaderAsIntOrDefault(String headerName, Object rawValue)` helper and parse the two headers **independently** (`flag` first, then `delayLevel`). A malformed value in one header no longer drags the other, well-formed header down to the fallback, and the warning names only the header that actually failed.
- The helper catches `NumberFormatException`, logs a `warn` with the header name, the raw value, and the exception **message only** (no stack trace, to avoid flooding logs on the per-message producer path), and falls back to `0`.

Difference from the source PR, per the review request: `2025.0.x` does not use `@NullMarked`, NullAway, or JSpecify, so the `org.jspecify.annotations.Nullable` import and annotation are **omitted**. Everything else (independent header parsing, warning behavior, regression tests) is kept identical.

### Describe how to verify it

`mvn -pl spring-cloud-alibaba-starters/spring-cloud-starter-stream-rocketmq test` — all 13 module tests pass locally, including the 6 tests carried over from #4330 in `RocketMQMessageConverterSupportTest`:

- `nonNumericDelayTimeLevelHeaderFallsBackToZero` / `nonNumericFlagHeaderFallsBackToZero` — malformed header falls back to `0`.
- `invalidNumericHeaderDoesNotPropagateException` — no exception escapes `convertMessage2MQ`.
- `nonNumericFlagHeaderLogsWarning` — `OutputCaptureExtension` asserts the warning contains the header name and raw value.
- `validDelayLevelWithNonNumericFlagIsPreservedAndWarnsOnlyFlag` / `validFlagWithNonNumericDelayLevelIsPreservedAndWarnsOnlyDelayLevel` — mixed-validity cases: the well-formed header keeps its value, the warning names only the failed one.

### Special notes for reviews

Same two intentional behaviour changes as #4330 (see its "Special notes" for the full rationale):

1. **Mixed-validity fallback changed (improvement).** Each header is parsed independently, so a well-formed header keeps its value even when the other one is malformed.
2. **Catch narrowed from `Exception` to `NumberFormatException` (intentional).** An exception thrown by a custom header value whose `toString()` throws now propagates instead of being silently converted into `flag=0`/`delayLevel=0`; `NumberFormatException` (the only expected failure) is still handled with warn-and-fallback.

### DIFF
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-stream-rocketmq/src/main/java/com/alibaba/cloud/stream/binder/rocketmq/support/RocketMQMessageConverterSupport.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-stream-rocketmq/src/main/java/com/alibaba/cloud/stream/binder/rocketmq/support/RocketMQMessageConverterSupport.java
@@ -26,6 +26,8 @@ import com.alibaba.cloud.stream.binder.rocketmq.convert.RocketMQMessageConverter
 import com.alibaba.cloud.stream.binder.rocketmq.custom.RocketMQBeanContainerCache;
 import org.apache.rocketmq.common.message.MessageConst;
 import org.apache.rocketmq.common.message.MessageExt;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageHeaders;
@@ -39,6 +41,9 @@ import org.springframework.util.ObjectUtils;
  * @author zkzlx
  */
 public final class RocketMQMessageConverterSupport {
+
+	private static final Logger log = LoggerFactory
+			.getLogger(RocketMQMessageConverterSupport.class);
 
 	private RocketMQMessageConverterSupport() {
 	}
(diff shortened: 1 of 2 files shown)


uuuyuqi (approved):
LGTM

PR #4385, opened 2026-08-26
fix(nacos): match YAML array keys during config refresh (2025.0.x backport of #4254)

### LINKED ISSUE #4239: Can not refresh when config using array in yaml

uuuyuqi:

Backports #4254 to `2025.0.x`.

#4239 was reported against `spring-alibaba-nacos-config:2025.0.0.0`. For YAML array/list values, the config change parser produces indexed keys such as `storeCodes[3]`, while `@NacosConfigKeysListener` usually registers the base key `storeCodes`. Exact matching therefore prevents these changes from triggering the listener. The fix was merged into `2025.1.x` through #4254 and is still missing from `2025.0.x`.

NONE (backport of #4254; related to the already closed #4239)

### Describe how you did it

- Keep exact interested-key matching as the first check.
- When exact matching fails, strip the first array-index suffix from keys such as `storeCodes[3]` or `config.items[0].name` and compare the resulting base key.
- Preserve the existing prefix matching and empty-filter behavior.
- Add six unit tests covering exact matching, indexed keys, nested paths, unrelated keys, prefix matching, and empty filters.

### Describe how to verify it

```bash
./mvnw -pl spring-cloud-alibaba-starters/spring-alibaba-nacos-config -am -Dtest=NacosPropertiesKeyListenerTest -Dsurefire.failIfNoSpecifiedTests=false test
./mvnw -pl spring-cloud-alibaba-starters/spring-alibaba-nacos-config -am test
```

Both commands pass locally. The focused test class runs 6 tests, and the complete Nacos Config module suite runs 20 tests with no failures or errors.

### Special notes for reviews

The production-code change and tests match #4254. The test copyright header uses `2013-2023` instead of `2013-present` to comply with the Checkstyle rules on the `2025.0.x` branch.

### DIFF
--- a/spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/annotation/NacosPropertiesKeyListener.java
+++ b/spring-cloud-alibaba-starters/spring-alibaba-nacos-config/src/main/java/com/alibaba/cloud/nacos/annotation/NacosPropertiesKeyListener.java
@@ -48,7 +48,7 @@ public abstract class NacosPropertiesKeyListener extends AbstractConfigChangeLis
 		if (CollectionUtils.isNotEmpty(interestedKeys) || CollectionUtils.isNotEmpty(interestedKeyPrefixes)) {
 			boolean foundInterested = false;
 			for (ConfigChangeItem changeItem : event.getChangeItems()) {
-				if (interestedKeys != null && interestedKeys.contains(changeItem.getKey())) {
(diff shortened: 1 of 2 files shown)


xuxiaowei-com-cn (approved):
LGTM

PR #4386, opened 2026-08-26
fix(nacos): avoid false cross-cluster warning when clusterName is unset (2025.0.x backport of #4287)

uuuyuqi:
### Describe what this PR does / why we need it

This backports #4287 to `2025.0.x`.

When `spring.cloud.nacos.discovery.cluster-name` is not configured, `NacosLoadBalancer` currently logs `A cross-cluster call occurs` for normal load-balanced requests. A blank cluster name does not represent a cross-cluster fallback, so the warning is misleading and can generate excessive logs.

The warning should only be emitted when a cluster name is configured but no same-cluster instance is available.

### Does this pull request fix one issue?

Fixes #4020

### Describe how you did it

- Moved the warning into the branch that handles a configured cluster with no matching instances.
- Throttled repeated cross-cluster warnings to at most once every 10 seconds per load balancer instance.
- Added regression tests for null and empty cluster names, same-cluster selection, cross-cluster fallback, warning throttling, and empty instance lists.
- Kept the `2025.0.x` branch-specific exception response behavior and copyright header format.

### Describe how to verify it

Run:

```bash
./mvnw \
  -pl spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery \
  -am \
  -Dtest=NacosLoadBalancerTests \
  -Dsurefire.failIfNoSpecifiedTests=false \
  test
```

Result:

```text
Tests run: 6, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

The Maven lifecycle also reported zero Checkstyle violations, and `git diff --check` passed.

### Special notes for reviews

This is a focused backport of #4287. Compared with the `2025.1.x` implementation, this change retains the existing `2025.0.x` exception-path response because #4288 is outside the scope of this backport, and it uses the copyright header format required by `2025.0.x`.

PR #4384, opened 2026-08-26
fix(nacos): mask access and secret keys in config logs (2025.0.x backport of #4244)

uuuyuqi:
### Describe what this PR does / why we need it

Backports #4244 to `2025.0.x`.

#4242 reported that exception and log messages could expose plaintext Nacos `accessKey` and `secretKey` values through `NacosConfigProperties#toString()`. The fix was merged into `2025.1.x` through #4244, while `2025.0.x` still prints both values without masking.

### Does this pull request fix one issue?

NONE (backport of #4244; related to the already closed #4242)

### Describe how you did it

- Add a `mask` helper that preserves `null` and empty values and replaces non-empty values with `******`.
- Use the helper for `accessKey` and `secretKey` in `NacosConfigProperties#toString()`.
- Add a unit test that verifies both credentials are masked and their original values do not appear in the output.

### Describe how to verify it

```bash
./mvnw -pl spring-cloud-alibaba-starters/spring-alibaba-nacos-config -am -Dtest=NacosConfigPropertiesTest -Dsurefire.failIfNoSpecifiedTests=false test
./mvnw -pl spring-cloud-alibaba-starters/spring-alibaba-nacos-config -am test
```

Both commands pass locally. The focused test runs 1 test, and the complete Nacos Config module suite runs 15 tests with no failures or errors.

### Special notes for reviews

The production-code change and test match #4244. The test copyright header uses `2013-2023` instead of `2013-present` to comply with the Checkstyle rules on the `2025.0.x` branch.

PR #4383, opened 2026-08-26
fix(nacos): deregister instances on DOWN status (2025.0.x backport of #4219)

uuuyuqi:
### Describe what this PR does / why we need it

Backports #4219 to `2025.0.x`.

#4208 was reported against Spring Cloud Alibaba `2025.0.0`. Calling the `ServiceRegistryEndpoint` with `status=DOWN` registered the same Nacos instance with `enabled=false`, but the instance remained online in Nacos. The fix was merged into `2025.1.x` through #4219 and is still missing from `2025.0.x`.

### Does this pull request fix one issue?

NONE (backport of #4219; related to the already closed #4208)

### Describe how you did it

- Delegate `DOWN` status handling to `NacosServiceRegistry#deregister`.
- Delegate `UP` status handling to `NacosServiceRegistry#register`.
- Add focused regression tests for both status transitions.

### Describe how to verify it

```bash
./mvnw -pl spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery -am -Dtest=NacosServiceRegistryTests -Dsurefire.failIfNoSpecifiedTests=false test
./mvnw -pl spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery -am test
```

Both commands pass locally. The focused test class runs 2 tests, and the complete Nacos Discovery module suite runs 50 tests with no failures or errors.

### Special notes for reviews

The production-code change matches #4219. This backport additionally adds focused regression coverage for the `DOWN` and `UP` paths.

PR #4387, opened 2026-08-26
fix(nacos): return EmptyResponse instead of null on exception (2025.0.x backport of #4288)

uuuyuqi:

This backports #4288 to `2025.0.x`.

When instance selection throws inside `NacosLoadBalancer.getInstanceResponse()`, the method currently returns `null`. This violates the `Response<ServiceInstance>` contract and can cause a downstream `NullPointerException` when callers invoke methods such as `hasServer()`.

The empty-instance path in the same method already returns `EmptyResponse`, so the exception path should return the same non-null response type.

NONE. The original fix was found during code review and merged through #4288 without a separate issue.

### Describe how you did it

- Changed the exception path from `return null` to `return new EmptyResponse()`.
- Added regression tests for the private instance-selection method and the public reactive `choose()` API.
- Covered exception handling, normal selection, and empty instance lists.
- Adapted the new test file to the copyright header format required by `2025.0.x`.

### Describe how to verify it

Focused regression tests:

```bash
./mvnw \
  -pl spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery \
  -am \
  -Dtest=NacosLoadBalancerErrorHandlingTests \
  -Dsurefire.failIfNoSpecifiedTests=false \
  test
```

Result:

```text
Tests run: 5, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

Complete affected module:

```bash
./mvnw \
  -pl spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery \
  -am \
  test
```

Result:

```text
Tests run: 55, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

The Maven lifecycle reported zero Checkstyle violations, and `git diff --check` passed.

### Special notes for reviews

- The backport is based on `2025.0.x` commit `a97ba40e9`, which includes #4383, #4384, and #4385.
- The code change matches #4288; only the test copyright header was adapted for `2025.0.x`.
- PR #4386 changes cross-cluster warning behavior in the same method, but this PR changes only the exception return value and remains independent.

### DIFF
--- a/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/main/java/com/alibaba/cloud/nacos/loadbalancer/NacosLoadBalancer.java
+++ b/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-discovery/src/main/java/com/alibaba/cloud/nacos/loadbalancer/NacosLoadBalancer.java
@@ -179,7 +179,7 @@ public class NacosLoadBalancer implements ReactorServiceInstanceLoadBalancer {
 		}
 		catch (Exception e) {
 			log.warn("NacosLoadBalancer error", e);
-			return null;
+			return new EmptyResponse();
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


xuxiaowei-com-cn (approved):
LGTM

